PR #40, opened 2014-04-21
Add address and port to socket error message for better troubleshooting.

Xangis:
Add address and port to socket error message in AMQP/transport.py so that it's clearer what we've failed to connect to.

I've found this helpful in troubleshooting connections and misconfigurations.

### DIFF
--- a/amqp/transport.py
+++ b/amqp/transport.py
@@ -92,7 +92,7 @@ class _AbstractTransport(object):
 
         if not self.sock:
             # Didn't connect, return the most recent error message
-            raise socket.error(last_err)
+            raise socket.error(last_err, 'Address: ' + host + ', Port: ' + str(port))
 
         try:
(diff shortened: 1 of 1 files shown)


ms7s:
This is something that would be helpful, I was about to write a similar patch when I found this pull request. Let me know if I should create a separate issue or pull request. Two observations:

- The exception arguments in master branch are not what the calling code might expect already, since `last_err` is itself an instance of `socket.error` (instead of a tuple of error code/message). This also means that the raised exception's `errno` attribute does not contain correct error number
- `socket.error` inherits from `IOError` since Python 2.6, so we might put the host name and port into `filename` attribute of the exception, which would include it in the error message (but I'm not sure whether abusing this attribute is a good idea, maybe creating a subclass of `socket.error` would work better)

auvipy:
open a new PR with suggested changes

PR #54, opened 2014-10-19
Remove redefine Transport method in Connection class

zhurongze:


ask:
Thanks! Seems like this was already fixed somehow

PR #55, opened 2014-10-19
add keepalive options

zhurongze:
if the rabbitmq-server's host crush, the consumer will not receive
'socket close' until 7200 seconds elapsed (os default value), so the
consumer cann't receive any message while in 7200 seconds.

the patch add keealive options, so we can set short keepalive_idle to
avoid it.


ask:
Thanks! Merged into master!

Issue #46, opened 2014-07-13
Cannot write frames or deserialize on Py3.4


jbg:
On Py3.4, it is necessary to apply the following patch in order to write frames or deserialize data:

```
diff --git a/amqp/method_framing.py b/amqp/method_framing.py
index b9ffe91..64e9cd4 100644
--- a/amqp/method_framing.py
+++ b/amqp/method_framing.py
@@ -116,7 +116,7 @@ def frame_writer(connection, transport,
 
         if no_pybuf or bigbody:
             # ## SLOW: string copy and write for every frame
-            frame = (''.join([pack('>HH', *method_sig), args])
+            frame = (b''.join([pack('>HH', *method_sig), args])
                      if type_ == 1 else '')  # encode method frame
             framelen = len(frame)
             write(pack('>BHI%dsB' % framelen,
diff --git a/amqp/serialization.py b/amqp/serialization.py
index d74515f..81abf3d 100644
--- a/amqp/serialization.py
+++ b/amqp/serialization.py
@@ -52,7 +52,7 @@ ILLEGAL_TABLE_TYPE = """\
 
 
 def _read_item(buf, offset=0, unpack_from=unpack_from):
-    ftype = buf[offset]
+    ftype = chr(buf[offset])
     offset += 1
 
     # 'S': long string
```

In method_framing.py, `''.join()` cannot be applied on a list of bytes, and in serialization.py, `buf[offset]` is an `int` but is compared to `str` a few lines below.

Note this patch will probably break Py2 so should not be applied verbatim, it's just to demonstrate the issue.

underyx:
Is it possible that #66 is caused by this issue?

ask:
Fixed in master!

PR #56, opened 2014-10-24
Python 3 fixes

pashinin:
These changes were already mentioned in https://github.com/celery/py-amqp/issues/46
It should work with Python 3 now


ask:
Thanks!  Sadly, I fixed this in master, without checking if there was a patch :(

PR #58, opened 2014-12-03
AbstractTransport overwrites user defined timeout

andronat:
Hello, the **AbstractTransport** is overwriting the user defined argument: **connect_timeout**, which leads to creating a socket with no timeout settings.

### DIFF
--- a/amqp/transport.py
+++ b/amqp/transport.py
@@ -95,7 +95,6 @@ class _AbstractTransport(object):
             raise socket.error(last_err)
 
         try:
-            self.sock.settimeout(None)
             self.sock.setsockopt(SOL_TCP, socket.TCP_NODELAY, 1)
             self.sock.setsockopt(socket.SOL_SOCKET, socket.SO_KEEPALIVE, 1)
 


andronat:
Any news on the subject?

dallasmarlow:
@ask it would be great to get this merged if possible, are there any blockers to this change?

dallasmarlow:
@andronat it looks like python's `socket.settimeout` changes the socket to a non-blocking mode which is not desirable for normal read/write operations here. this library uses `socket.settimeout` to provide the `connect_timeout` functionality only and then sets python's socket timeout value back to `None` in order to change the socket mode back to blocking meaning that this PR is probably not what you want.

dallasmarlow:
@andronat https://github.com/celery/py-amqp/pull/64 provides functional socket timeouts (although you'll need to set both read and write timeouts to get the desired effect)

andronat:
@dallasmarlow ok I see. Should I cancel this PR?

dallasmarlow:
@andronat It's up to you, but I think that's probably best.

andronat:
I close this issues is solved with #64.

PR #62, opened 2015-03-10
Adding ability to pass an ssl context to amqp ssl transport.

vickumar1981:
Allows users to create a ssl context and pass that context to the AMQP SSL Transport layer.  This allows a user to set custom SSL options for the connection, such as whether they'd like the certificate's host name to be verified, and whether a certificate is required or not.

    -  Allows passing a 'context' attribute to the amqp transport, to override the default context
    -  If context is not present, falls back to normal functionality of using ssl.wrap_socket()


vickumar1981:
Hey, we have some fixes to iron out.  Please do not merge, yet.

PR #63, opened 2015-03-16
Add option to check host name in amqp ssl transport.

vickumar1981:
Allows users to create a default ssl context and pass options to that context in the AMQP SSL Transport layer.  This allows a user to set custom SSL options for the AMQP connection, such as whether they'd like the certificate's host name to be verified, and whether a certificate is required or not.

    -  Allows passing a 'context' attribute to the amqp transport, to override settings for SSL.wrap_socket()
    -  If context is not present, falls back to normal functionality of using ssl.wrap_socket()


ask:
Thanks! Merged into master

PR #64, opened 2015-03-26
add blocking socket snd/rcv timeouts as transport options

dallasmarlow:


andronat:
@dallasmarlow thanks again for solving this. Just one commend if I may, can you add a commend on the  `self.sock.settimeout(None)` that this is needed to overcome non-blocking behavior? So the next person that sees the code will not have to image what's going on.

dallasmarlow:
@andronat np, i've added a comment. =)

dallasmarlow:
@ask would you mind reviewing this pr?

amqp/transport.py
```diff
@@ -54,7 +55,7 @@ class _AbstractTransport(object):
     """Common superclass for TCP and SSL transports"""
     connected = False
 
-    def __init__(self, host, connect_timeout):
+    def __init__(self, host, connect_timeout, write_timeout, read_timeout):
```
FlaPer87: In L105 you check if `write_timeout` and `read_timeout` are None. I think it'd be worth having these 2 options as keyword arguments instead of being required.

Also, without this, we'd break backwards compatibility for consumers of py-amqp

dallasmarlow: @FlaPer87 i was under the impression that most consumers of this lib were using the connection class instead of the transport classes directly, i'll look into updating this pr to make those optional arguments

FlaPer87: @dallasmarlow actually, you're right. It is marked as private. However, it feels weird that we require it in the constructor signature but we check if it is None later on. I guess making it a keyword is not bad.

dallasmarlow: @FlaPer87 now that i'm looking at this again, i remember that argument ordering for `SSLTransport` and `create_transport` will be a bit odd if we want to make everything backwards compatible internally. i've pushed a branch on my fork (https://github.com/dallasmarlow/py-amqp/pull/2) which has those changes, let me know if you would like to include those changes here.

FlaPer87: @dallasmarlow yup, it seems that we could have those changes here. Thanks!

dallasmarlow:
@FlaPer87 this pr has been updated to preserve internal backwards compatibility as requested

FlaPer87:
@dallasmarlow just one more thing and I'll merge it. Would you mind squashing those commits? I don't think they bring much info to the history. I'd like this to be a single change.

dallasmarlow:
@FlaPer87 my fork has now been rebased to include only one commit

FlaPer87:
@dallasmarlow it looks like there's a merge commit now. Mind squashing, I'll approve after that. No idea what happend.

dallasmarlow:
@FlaPer87 i forked off of the last released version to make a minimal patch to solve the timeout issue. when i rebased master from my fork as you requested, git was no longer able to merge https://github.com/celery/py-amqp/commit/7ef76ecde592dd7043b0f3703de7e22e2211f15c into my changes and the diff showed the old value as part of my pr. i then pulled this repo's changes into my fork which created the merge commit, and now i have rebased again to remove it as requested.

FlaPer87:
@dallasmarlow :cake: there you go, thanks!

dims:
@FlaPer87 @dallasmarlow thanks!

PR #65, opened 2015-04-30
Allow to set a default timeout to the socket

sileht:
Hi,

Propose change to help to fix: https://github.com/celery/kombu/issues/463

Cheers,

### DIFF
--- a/amqp/connection.py
+++ b/amqp/connection.py
@@ -143,7 +143,7 @@ class Connection(AbstractChannel):
                  ssl=False, connect_timeout=None, channel_max=None,
                  frame_max=None, heartbeat=0, on_open=None, on_blocked=None,
                  on_unblocked=None, confirm_publish=False,
-                 on_tune_ok=None, **kwargs):
+                 on_tune_ok=None, client_timeout=None, **kwargs):
         """Create a connection to the specified host, which should be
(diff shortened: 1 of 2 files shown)


michaelklishin:
Thank you for looking into this.

Note that per spec the client must consider 2 skipped heartbeats to be a (heartbeat) timeout.

Some clients send heartbeats at half one interval and set socket's read timeout to the same value. Then 4 socket timeouts is considered a failure. This helps with eliminating race condition effects with lower timeouts (e.g. 3-6 seconds), so may be an overkill for `py-amqp`.

dallasmarlow:
@sileht i believe that this change will not accomplish your goal, see https://github.com/celery/py-amqp/pull/58#issuecomment-86488765 which is another pr which proposed passing a user defined value to `settimeout`. i believe that https://github.com/celery/py-amqp/pull/64 will add the functionality you would expect from such an option.

dallasmarlow:
@michaelklishin one thing worth mentioning is that in the latest released version of this lib (`1.4.6`), heartbeat requests can block forever which makes implementing a system which uses them a bit of extra work. heartbeat requests using the a library built from the head of master will timeout using an interval from the `connect_timeout` connection class attribute.

sileht:
Thanks @dallasmarlow  #64 should fix my issue.

PR #69, opened 2015-08-19
Fix issue with remote connection hung w/ SSL transport

aguinet:
There was a test (but misplaced!) on an exception potentially thrown by
the python SSL library. Python 3.4 and 3.2 have different behaviours, so
both are checked!


malinoff:
Thanks!

PR #70, opened 2015-10-16
Reset self.channels to emtpy dict instead of None when closing connection.

apporc:
1. The initial value of self.channels is {}, not None.

2. After Connection.close() was invoked, if self.channels set to None,
invoking Connection.channel() will raise unexpected error.

In such case:

    try:
        return self.channels[channel_id]
    except KeyError:
        return self.Channel(self, channel_id, on_open=callback)

code above will raise `TypeError: 'NoneType' object has no attribute '__getitem__'`,
which will not got cought.

Signed-off-by: apporc <appleorchard2000@gmail.com>

### DIFF
--- a/amqp/connection.py
+++ b/amqp/connection.py
@@ -333,7 +333,8 @@ class Connection(AbstractChannel):
         except socket.error:
             pass  # connection already closed on the other end
         finally:
-            self.transport = self.connection = self.channels = None
+            self.transport = self.connection = None
+            self.channels = {}
 
     def _get_free_channel_id(self):
         try:


ask:
What happens after this change though? It shouldn't be possible to create a new channel after the connection is closed

apporc:
After this , when the code try to create a new channel. A connection error will be generated, so the one who tried to create the channel be able to know  the ptoblem, then it can choose to do something else to recover the conbection or just quit normally.. But if just Nonetype error generated, the program crashes.

ask:
I feel more safe keeping it at None, and we want it to raise RecoverableConnectionError in this case

PR #71, opened 2015-10-19
Add 'nowait' to Connection.close(), buffer incoming data

smurfix:
Buffering: calling read() for a few bytes is significantly more expensive than simply reading everything at once.

.close(): After setting up the channels, running drain_events() in a separate thread works quite well – except there's no way to cleanly close the connection.

### DIFF
--- a/amqp/abstract_channel.py
+++ b/amqp/abstract_channel.py
@@ -52,7 +52,7 @@ class AbstractChannel(object):
 
     def send_method(self, sig,
                     format=None, args=None, content=None,
-                    wait=None, callback=None):
+                    wait=None, callback=None, returns_tuple=False):
         p = promise()
         conn = self.connection
         if conn is None:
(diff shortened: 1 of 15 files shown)


ask:
Could you clean this up into multiple pull requests please? And why do we want to make the channel thread-safe?

smurfix:
On 27.10.2015 19:42, Ask Solem Hoel wrote:
>
> Could you clean this up into multiple pull requests please?
>
Will do.
>
> And why do we want to make the channel thread-safe?
>
Think of a RPC system where the server spawns separate threads (or uses
a thread pool) to process incoming requests. Obviously, when two replies
become ready at the same time, one of them will have to wait.

However, I've decided to switch my own code to asyncio. I have already
managed to re-implement amqp's Promise class on top of asyncio.Future;
the work getting the functional tests to pass again is about half done.
With asyncio, locking at that level isn't necessary since there is no
implicit task switching, so I'll drop that part.

-- 
-- Matthias Urlichs

thedrow:
This PR has gone stale. If you want to continue working on it please open a new PRs containing small enough changes for us to review.

PR #75, opened 2015-12-19
Check for ctypes on OSX and Linux

lewisjared:
When using py-amqp and Google App Engine (GAE), an ImportError is raised when importing ctypes. This patch enables the use of py-amqp on GAE.

A similar fix has been applied to kombu.


ask:
Thanks! Merged into master and 1.4

Issue #76, opened 2015-12-31
struct.error: argument for 's' must be a bytes object


nszceta:
"amqp/method_framing.py", line 161, in frame_writer
    3, channel, framelen, body, 0xce)
struct.error: argument for 's' must be a bytes object

PR #77, opened 2015-12-31
Fix Issue #76

nszceta:


amqp/method_framing.py
```diff
@@ -157,6 +157,10 @@ def frame_writer(connection, transport,
                 offset += 8 + framelen
 
                 framelen = len(body)
+
+                if isinstance(body, str):
+                    body = body.encode()
+
```
ask: I hate Python 3 for requiring me to unnecessarily copy this string!   There should be a way for the new bytes object to use the strings underlying C array

ask: At least we want a separate branch for Python 3 here, so this behavior doesn't leak into our beloved Python 2.

PR #81, opened 2016-03-31
checked for one more version of librt.so

mfahadsid:
there was only check for one version of librt.so.x ... according to my case as im using this client in some embedded systems. i think there should be some other version checks too. like librt.so.0 (worked for me) ... or latest versions.


ask:
Thanks!  This part of the code has since been moved to celery/vine, but I tried to preserve the commit information as much as possible:

celery/vine@f85ceec5f2848dd706b9c697e3208260bc65d92b

mfahadsid:
You're welcome.

PR #87, opened 2016-05-27
fix pack/unpack argument unicode error for python version < 2.7.6

air-upc:


ask:
Thanks!  Seems this was fixed in a different pull request also that I already merged: e69b20323ee74d5b63a688d6f875120fffed474f

Issue #85, opened 2016-05-27
ImportError: cannot import name spec


MrKiven:
Hi, we have got this error

``ImportError: cannot import name spec``

we use `librabbitmq==1.6.1`, `kombu==3.0.33`

Traceback here:

``
File "/srv/ves/arch.note/.venv/lib/python2.7/site-packages/kombu/__init__.py", line 67, in __getattr__

module = __import__(object_origins[name], None, None, [name])

File "/srv/ves/arch.note/.venv/lib/python2.7/site-packages/kombu/connection.py", line 19, in <module>

from kombu import exceptions

File "/srv/ves/arch.note/.venv/lib/python2.7/site-packages/kombu/exceptions.py", line 12, in <module>

from amqp import ChannelError, ConnectionError, ResourceError

File "/srv/ves/arch.note/.venv/lib/python2.7/site-packages/amqp/__init__.py", line 39, in <module>

from .basic_message import Message # noqa

File "/srv/ves/arch.note/.venv/lib/python2.7/site-packages/amqp/basic_message.py", line 19, in <module>

from . import spec

ImportError: cannot import name spec
``

air-upc:
Before Python version 2.7.7, the struct.pack() function required a native string as its format argument.

MrKiven:
anyone can check it? @ask

bastbnl:
I'm seeing the same message using python 2.7.3 using amqp==2.0.0 since I've upgraded amqp-1.4.9 this morning.

...
from amqp import ChannelError, ConnectionError, ResourceError
  File "/var/www/brownpapersession/dev/env-brownpapersession/local/lib/python2.7/site-packages/amqp/__init__.py", line 39, in <module>
    from .basic_message import Message      # noqa
  File "/var/www/brownpapersession/dev/env-brownpapersession/local/lib/python2.7/site-packages/amqp/basic_message.py", line 19, in <module>
    from . import spec
ImportError: cannot import name spec

ask:
Hey!

Kombu 3.x should not be used with amqp 2.0, and the kombu requirements have reflected this since the release of 3.0.0:
https://github.com/celery/kombu/blob/3.0/requirements/default.txt

If you use kombu 3.x, you should keep to the 1.4 series of py-amqp!

ask:
That said, I don't know if kombu 3.x is actually incompatible, and I don't understand why this error is happening.

[...]

PR #88, opened 2016-05-27
Fixes issue #85: the module can be imported on python 2.7.3 running i…

bastbnl:
…n unicode mode. I guess sys.setdefaultencoding('utf-8') breaks the arguments to the pack


ranman:
Is this pushed to pypi yet?

MrKiven:
great...

Issue #91, opened 2016-06-01
openstack oslo.messaging load failed due to amqp can't find module 'vine' in Python3.4


glongwave:
Traceback (most recent call last):
  File "/home/jenkins/workspace/gate-smaug-python34/smaug/tests/base.py", line 92, in setUp
    self.messaging_conf = messaging_conffixture.ConfFixture(CONF)
  File "/home/jenkins/workspace/gate-smaug-python34/.tox/py34/lib/python3.4/site-packages/oslo_messaging/conffixture.py", line 50, in __init__
    'oslo_messaging_rabbit')
  File "/home/jenkins/workspace/gate-smaug-python34/.tox/py34/lib/python3.4/site-packages/oslo_messaging/conffixture.py", line 25, in _import_opts
    __import__(module)
  File "/home/jenkins/workspace/gate-smaug-python34/.tox/py34/lib/python3.4/site-packages/oslo_messaging/_drivers/impl_rabbit.py", line 30, in <module>
    import kombu.connection
  File "/home/jenkins/workspace/gate-smaug-python34/.tox/py34/lib/python3.4/site-packages/kombu/connection.py", line 19, in <module>
    from kombu import exceptions
  File "/home/jenkins/workspace/gate-smaug-python34/.tox/py34/lib/python3.4/site-packages/kombu/exceptions.py", line 12, in <module>
    from amqp import ChannelError, ConnectionError, ResourceError
  File "/home/jenkins/workspace/gate-smaug-python34/.tox/py34/lib/python3.4/site-packages/amqp/__init__.py", line 39, in <module>
    from .basic_message import Message # noqa
  File "/home/jenkins/workspace/gate-smaug-python34/.tox/py34/lib/python3.4/site-packages/amqp/basic_message.py", line 29, in <module>
    from .serialization import GenericContent
  File "/home/jenkins/workspace/gate-smaug-python34/.tox/py34/lib/python3.4/site-packages/amqp/serialization.py", line 34, in <module>
    from .exceptions import FrameSyntaxError
  File "/home/jenkins/workspace/gate-smaug-python34/.tox/py34/lib/python3.4/site-packages/amqp/exceptions.py", line 21, in <module>
    from .five import python_2_unicode_compatible
  File "/home/jenkins/workspace/gate-smaug-python34/.tox/py34/lib/python3.4/site-packages/amqp/five.py", line 14, in <module>
    import vine.five
ImportError: No module named 'vine'

liu-sheng:
[...]

PR #92, opened 2016-06-02
Fix can't install requirements in Python 3

glongwave:
Close https://github.com/celery/py-amqp/issues/91


ask:
Thanks!  Merged into master

PR #93, opened 2016-06-02
Fix unicode literals (Python 2.7)

alanjds:
The existing version breaks on Python 2.7 because `struct()` needs bytes as the 1st parameter, not unicode.


alanjds:
Someone please reload appveyor. It is broken. (maybe) not the tests.

ask:
This is on Python 2.7.3 only I believe.

I'm not sure why they are fixing these bugs in the 2.7 series, it's crazy as it means we have to test against every single release version of Python, when they could just have released Python 2.8

ask:
Thanks! Merged into master

PR #96, opened 2016-06-03
patch to fix struct pack/unpack unicode error when python version < 2.7.7

air-upc:
patch to fix "TypeError: Struct() argument 1 must be string, not unicode" when python version < 2.7.7

### DIFF
--- a/amqp/exceptions.py
+++ b/amqp/exceptions.py
@@ -16,9 +16,8 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301
 from __future__ import absolute_import, unicode_literals
 
-from struct import pack, unpack
-
 from .five import python_2_unicode_compatible
+from .utils import pack, unpack
 
 __all__ = [
     'AMQPError',
@@ -260,5 +259,4 @@ METHOD_NAME_MAP = {
 
 
 for _method_id, _method_name in list(METHOD_NAME_MAP.items()):
-    METHOD_NAME_MAP[unpack(u'>I', pack(u'>HH', *_method_id))[0]] = \
-        _method_name
+    METHOD_NAME_MAP[unpack('>I', pack('>HH', *_method_id))[0]] = _method_name
--- a/amqp/method_framing.py
+++ b/amqp/method_framing.py
@@ -17,13 +17,12 @@
 from __future__ import absolute_import, unicode_literals
 
 from collections import defaultdict
-from struct import pack, unpack_from, pack_into
 
 from . import spec
 from .basic_message import Message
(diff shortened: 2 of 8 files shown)


air-upc:
https://hg.python.org/cpython/raw-file/f89216059edf/Misc/NEWS #19099
We have got burned as py-amqp not works with python 2.7.5; will you make it compatible with python < 2.7.7? any solution is OK, not restricted to this pull request... @ask

ask:
I just merged #93, which looks like fixes the same.

I do like this version though, as it's cleaner and only degrades performance on old Python versions,
so maybe we can use this instead.

air-upc:
yeah, #93 fix the same problem, and I am so glad that we don't need to upgrade our python version in the near future as it's much exhausted.

alanjds:
I did not knew that there are specific behaviours for different 2.7.X versions. It is a shame.

Anyway I do prefer this #96 over mine, as this concentrates the "compat" stuff in one place.

ask:
Both of them are good really!  This version in #96 will add some overhead when using older Python versions.  I want pyamqp to be as fast as librabbitmq, and method calls/object lookups are exactly the sort of stuff I've been optimizing away in celery to gain better benchmarks (it adds up to quite a lot of message/s in the end).  If it was required on later Python 2.7 versions (our main deployment base), I would take #93.

I will find the time to merge both soon

alanjds:
I was not aware of this performance target. Had you yet tested with "whole code" general optimisers/compilers like Numba, Nuitka or RPython yet? (just for me not try something already knew as a dead-end)

franklingu:
So will this be in the latest release? I am using amqp==2.0.3 and clearly it is broken on Python 2.7.3 now.

I am still using Ubuntu 12.04 so I am using 2.7.3 now

air-upc:
@ask  Sorry for haven't check this pull request for such a long time, but the coverage check result seems not good...

yujiezeng:
Hello, I have been use this project, with python version of 2.7.3. I have got this error too, but I have used the patch file. I think the problem happens in _trasport.py_, which slao use _unpack_ from _struct_. So I changed it  and it works well. Maybe I think you should have some modification on _trasnport.py_

air-upc:
@yujiezeng Modification to file `transport.py` lost when I rebase the pull request, what a terrible thing I have done. Thank you for the comment.

AndyMan1:
FYI Pull #93 didn't include exceptions.py, so I'm seeing the same error from this line:
https://github.com/celery/py-amqp/blob/master/amqp/exceptions.py#L263

It looks like this pull request does include that fix.

I'm using py-amqp 2.1.0 and Python 2.7.5.

PR #110, opened 2016-10-07
Refactored to make authn modular; added GSSAPI

alexsdutton:
This is an early pull request to achieve two things:
- make the authentication aspects more extensible
- implement GSSAPI authentication as an option

I'm not expecting it to be merged immediately, but it'd be good to have some initial feedback.

WRT the GSSAPI stuff, I've written a [rabbitmq module](https://github.com/ox-it/rabbitmq-auth-mechanism-gssapi) that it works with, but as far as I know that's the only server-side implementation (qpid does GSSAPI, but that's AMQP 1.0).

I also know that SASL is a potentially multi-step process, whereas this implementation only supports a single step.

### DIFF
--- a/amqp/connection.py
+++ b/amqp/connection.py
@@ -21,8 +21,6 @@ import socket
 import uuid
 import warnings
 
-from io import BytesIO
-
 from vine import ensure_promise
 
 from . import __version__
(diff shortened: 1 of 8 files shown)


alexsdutton:
I think this is now in a state where the code changes could be merged in, but the docs would still need updating. Could someone give it a look over?

auvipy:
thanks, but could you try to fix the conflicts?

alexsdutton:
@auvipy Done; just need to tidy up the newly-introduced pydocstyle failures (and then ever so slightly improve the test coverage)

ask:
Thanks! It's looking really good.  I guess we need to release this in a amqp 2.2 version

PR #112, opened 2016-11-07
Pass fmt string as a str to struct.pack() and struct.unpack()

jdufresne:
Fixes potential warning when running Python with -b flag:

BytesWarning: Comparison between bytes and string

[struct docs](https://docs.python.org/3/library/struct.html)

### DIFF
--- a/amqp/exceptions.py
+++ b/amqp/exceptions.py
@@ -296,5 +296,5 @@ METHOD_NAME_MAP = {
 
 
 for _method_id, _method_name in list(METHOD_NAME_MAP.items()):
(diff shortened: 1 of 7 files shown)


auvipy:
thanks jhon

ask:
Python <2.7.6 doesn't support passing unicode as the argument to struct though, which is why we used bytes

jdufresne:
> Python <2.7.6 doesn't support passing unicode as the argument to struct though, which is why we used bytes

I see. In that case, I believe it should always be `str`. Not always `bytes` nor always `unicode`. See https://github.com/python/typeshed/pull/669 for a discussion of why using bytes is wrong. By reverting this, this is again causing lots of `BytesWarning` on Python 3.

PR #114, opened 2016-11-21
Set TCP_USER_TIMEOUT if available for better failover

JelteF:
This commit uses the TCP_USER_TIMEOUT to drop the connection when no response
has been received from rabbitmq in the connect_timeout time. This allows for a
more responsive failover when the server dies.

It is slightly based on this code https://github.com/openstack/oslo.messaging/blob/9c25485adb661d44b7f3f0c676ac3705e5754796/oslo_messaging/_drivers/impl_rabbit.py#L938-L954

### DIFF
--- a/amqp/transport.py
+++ b/amqp/transport.py
@@ -17,10 +17,12 @@
 from __future__ import absolute_import, unicode_literals
 
 import errno
+import platform
 import re
 import struct
 import socket
 import ssl
+import sys
 
 from contextlib import contextmanager
 from struct import unpack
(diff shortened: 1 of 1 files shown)


JelteF:
See this for detais on what TCP_USER_TIMEOUT is: https://patchwork.ozlabs.org/patch/62889/

JelteF:
The code coverage problems are only caused by not going over OS specific branches.

auvipy:
thanks, could you also add some test?

JelteF:
This is best tested by using the iptables command during the the test to drop all the packets. I'm not sure how to do that in the best way here.

ask:
@JelteF I think @auvipy means unit tests, and unit tests never depends on any specific environment such as iptables.  These tests should actually never use the network at all, but use mocking instead.

ask:
From what I can tell in `/usr/include/linux/tcp.h` the constant for `TCP_USER_TIMEOUT` is 18, and not 19 as used here:

https://apt-browse.org/browse/debian/wheezy/main/i386/linux-libc-dev/3.2.78-1/file/usr/include/linux/tcp.h
```c
#define TCP_USER_TIMEOUT	18	/* How long for loss retry before timeout */
```

ask:
Merged into master, made some small changes so that TCP_USER_TIMEOUT can be part of the socket_settings.

JelteF:
Good call on the changes, not sure how it happened that I increased the constant by one. The only thing that is a bit weird now is that `user_timeout` is an argument to `_set_socket_options`, but gets overwritten if `self.connect_timeout` is set.

PR #117, opened 2016-12-08
Pass fmt string as a str object to struct functions (#112)

jdufresne:
@ask This is intended to fix the use of the `struct` module in a way that is compatible with both Python 2 and 3 and doesn't produce any spurious warnings.

In Python 3, mixing `bytes` and `str` as the format argument for `struct` functions produces `BytesWarning`s. See the typeshed discussion, <https://github.com/python/typeshed/pull/669>

Per the [docs](https://docs.python.org/3/library/struct.html#format-strings), the format arg is a string (`str`), not bytes.

### DIFF
--- a/amqp/exceptions.py
+++ b/amqp/exceptions.py
@@ -296,5 +296,5 @@ METHOD_NAME_MAP = {
 
 
 for _method_id, _method_name in list(METHOD_NAME_MAP.items()):
-    METHOD_NAME_MAP[unpack(b'>I', pack(b'>HH', *_method_id))[0]] = \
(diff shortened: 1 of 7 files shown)


ask:
It's really annoying that we have to copy the string, and call this function so many times for every frame, maybe we should rather override `unpack` + `pack` on Python <2.7.6 so they can have the penalty:

```
if sys.version_info < (2, 7, 6):

    def unpack_from(s, *args, **kwargs):
        return struct.unpack(s.encode(), *args, **kwargs)
```

ask:
Merged your diff into master, then applied the discussed changes :)

PR #121, opened 2016-12-20
do not reconnect on collect

chungg:
collect is called to cleanup when there is a connection error but
the backward compatibility code to connect when transport is called
gets triggered. this doesn't seem right as this should be handled by
ensure_connection.


auvipy:
isn't any test/verification needed for the changes?

PR #126, opened 2016-12-29
On Windows work around setsockopt error in transport.

wmw:
On Windows in transport TCP_OPTS will include TCP_KEEPIDLE which has a value of 4. Calls to getsockopt and setsockopt with level of IPPROTO_TCP (SOL_TCP) only accept TCP_NODELAY however, and instead interpret TCP_KEEPIDLE as SO_REUSEADDR. Calls to setsockopt with IPPROTO_TCP, SO_REUSEADDR will fail with Windows error 10042 WSAENOPROTOOPT. This change exposes WIN_VERSION in platform if appropriate and modifies transport to skip TCP_OPTS on Windows.


wmw:
Going to redo this with a change to KNOWN_TCP_OPTS instead.

PR #139, opened 2017-04-05
Add support for setting the SNI hostname header

dhananjaysathe:
Replaces the default method `ssl.wrap_socket` with a patched one that supports passing in the server_hostname field. 
The _wrap_socket_sni method is based on the `ssl.wrap_socket` helper function (https://github.com/python/cpython/blob/2.7/Lib/ssl.py#L931) but adds in the extra `server_hostname` parameter that sets the required SNI headers.

### DIFF
--- a/amqp/transport.py
+++ b/amqp/transport.py
@@ -293,12 +293,28 @@ class SSLTransport(_AbstractTransport):
     def _wrap_socket(self, sock, context=None, **sslopts):
         if context:
             return self._wrap_context(sock, sslopts, **context)
-        return ssl.wrap_socket(sock, **sslopts)
+        return self._wrap_socket_sni(sock, **sslopts)
 
     def _wrap_context(self, sock, sslopts, check_hostname=None, **ctx_options):
         ctx = ssl.create_default_context(**ctx_options)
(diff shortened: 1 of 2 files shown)


auvipy:
check the build failures

dhananjaysathe:
@auvipy the builds are failing (eg: https://travis-ci.org/celery/py-amqp/jobs/218784453) its seems to complain  AttributeError: 'module' object has no attribute 'PROTOCOL_TLS'
However, the 'ssl' module has an attribute PROTOCOL_TLS am i missing something here ?

dhananjaysathe:
Hey @auvipy all test now pass there is some negligible drop in codeconv . I tried my best to keep this minimal but a bulk of it does not seem to be related to the patch at all. Could you please look into it and have this merged

auvipy:
thanks

thedrow:
Why did you change the SSL protocol version? What if it's not supported on the platform?

dhananjaysathe:
@thedrow both are alias for int 2 as far as all implementations I checked. Some versions of python don't have ssl.PROTOCOL_TLS defined

PR #146, opened 2017-05-22
Release notes for 2.2.0

thedrow:
These are the release notes for 2.2.0. 
This release is going to be used for the upcoming Celery release.
The motivation behind the minor version bump is due to the authentication refactor which adds new capabilities but doesn't break backwards compatibility.

Please review and let me know if I got anything wrong.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -5,6 +5,45 @@ py-amqp is fork of amqplib used by Kombu containing additional features and impr
 The previous amqplib changelog is here:
 http://code.google.com/p/py-amqplib/source/browse/CHANGES
 
+.. _version-2.2.0
+
+2.2.0
+=====
+:release-date: TBD
+:release-by: Ask Solem
+
+- Fix random delays in task execusion.
+
+  This is a bug that caused performance issues due to polling timeouts that occur when receiving incomplete AMQP frames. (Issues #3978 #3737 #3814) 
+  
+  Fix contributed by **Robert Kopaczewski**
+
+- Calling ``conn.collect()`` multiple times will no longer raise an ``AttributeError`` when no channels exist.
+
+  Fix contributed by **Gord Chung**
+
+- Fix compatibility code for Python 2.7.6.
+
+  Fix contributed by **Jonathan Schuff**
+
+- When running in Windows, py-amqp will no longer use the unsupported TCP option TCP_MAXSEG.
+
+  Fix contributed by **Tony Breeds**
+  
+- Added support for setting the SNI hostname header.
+
+  The SSL protocol version is now set to SSLv23
+
+  Contributed by **Dhananjay Sathe**
+
+- Authentication mechanisms were refactored to be more modular. GSSAPI authentication is now supported.
+
+  Contributed by **Alexander Dutton**
+  
+- Do not reconnect on collect.
+
+  Fix contributed by **Gord Chung**
+
 .. _version-2.1.4:
 
 2.1.4
--- a/Changelog
+++ b/Changelog
@@ -12,10 +12,10 @@ http://code.google.com/p/py-amqplib/source/browse/CHANGES
 :release-date: TBD
 :release-by: Ask Solem
 
-- Fix random delays in task execusion.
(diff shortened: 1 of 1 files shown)


Changelog
```diff
@@ -5,6 +5,45 @@ py-amqp is fork of amqplib used by Kombu containing additional features and impr
...
 http://code.google.com/p/py-amqplib/source/browse/CHANGES
 
+.. _version-2.2.0
+
+2.2.0
+=====
+:release-date: TBD
+:release-by: Ask Solem
+
+- Fix random delays in task execusion.
```
georgepsarakis: Small type here: `execution`

Changelog
```diff
@@ -5,6 +5,45 @@ py-amqp is fork of amqplib used by Kombu containing additional features and impr
...
+
+  Fix contributed by **Jonathan Schuff**
+
+- When running in Windows, py-amqp will no longer use the unsupported TCP option TCP_MAXSEG.
+
+  Fix contributed by **Tony Breeds**
+  
+- Added support for setting the SNI hostname header.
+
+  The SSL protocol version is now set to SSLv23
```
georgepsarakis: I am wondering if this change is a secure and sane default. The documentation states that this constant will be [deprecated](https://docs.python.org/2/library/ssl.html#ssl.PROTOCOL_SSLv23) in 2.7.13 and is actually an alias for `ssl.PROTOCOL_TLS`.

michael-k:
I've noticed the same typo and same typo and some whitespace issues.  See PR #147

Changelog
thedrow: I'm wondering the same as well. Should we revert this and recreate a pull request?

thedrow:
@auvipy @georgepsarakis Should we revert and recreate the pull request for SNI support? Do we understand the implications of setting our SSL implementation to SSLv23?
I think it's the safest bet.

auvipy:
@thedrow go for it as it seems to be safest bet

georgepsarakis:
@thedrow I agree with recreating the SNI support PR.

dhananjaysathe:
Hey so I thought about the compatiblity. Infact you can pass a more specific version as well as a dict in the new code . Sslv23 is the most compatible version of found on a number of devices and architectectures.
https://docs.python.org/2/library/ssl.html#protocol-versions
Here is a more detailed matrix here
https://docs.python.org/2/library/ssl.html#socket-creation

If you look at python 3 protocoltls and sslv23 are alias. It defaults to protocoltls 
https://docs.python.org/3/library/ssl.html#socket-creation
https://docs.python.org/3/library/ssl.html#protocol-versions

Here is python 2.6 (I actually found protocoltls to break on an arm board with these old versions)
https://docs.python.org/2.6/library/ssl.html#functions-constants-and-exceptions

If you look at the source across all versions both these alias to integer value 2

That said give me a day or so I'll add more exception handling and checks in place so it has the same behavior across every combination.

thedrow:
Why do we have to set it? It wasn't set before and everything worked as expected.

dhananjaysathe:
Hello @thedrow and other kind folks 
The earlier code called the built-in `ssl.wrap_socket` method (https://github.com/celery/py-amqp/pull/139/files#diff-e3b09014ce12269dff5eece1f9f7c082L296) src here: this doesn't allow one to pass the needed headers https://github.com/python/cpython/blob/master/Lib/ssl.py#L1131

The workaround to this was manually creating a context dummy options (dict) for the context (https://github.com/dhananjaysathe/py-amqp/blob/4fec0db649679c6e04d3698949e28dd1dbb0527b/amqp/transport.py#L298)and passing it ssl_opts as well so the context.wrap_socket would be called that supported setting the headers. This is tad messy.  On the other hand, we could allow passing a ssl.SSLContext object directly. If you want a PR for that I could provide one. Lemme know what you folks think.

Also here is the version compatibility  line from CPython src:
https://github.com/python/cpython/blob/master/Lib/ssl.py#L153

also sent a PR with some more checks 
https://github.com/celery/py-amqp/pull/150

PR #149, opened 2017-05-25
Fix _wrap_socket_sni method definition

dhananjaysathe:
the previous PR breaks the method


dhananjaysathe:
@auvipy PR https://github.com/celery/py-amqp/pull/139 is flawed  this fixes the _wrap_socket_sni method definition

PR #150, opened 2017-05-29
SNI PR fixes 

dhananjaysathe:
1) Add some checks around before passing SNI headers
2) Fixes around ssl version set
Based on comments in https://github.com/celery/py-amqp/pull/146
 cc @auvipy @thedrow

### DIFF
--- a/amqp/transport.py
+++ b/amqp/transport.py
@@ -326,9 +326,9 @@ class SSLTransport(_AbstractTransport):
             opts['ssl_version'] = ssl_version
         else:
             if hasattr(ssl, 'PROTOCOL_TLS'):
-                opts['ssl_version'] = ssl.PROTOCOL_TLS # newer python 2.7 & 3.x
+                opts['ssl_version'] = ssl.PROTOCOL_TLS
             else:
-                opts['ssl_version'] = ssl.PROTOCOL_SSLv23 #python 2.6 & old 2.7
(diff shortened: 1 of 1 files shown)


amqp/transport.py
```diff
@@ -306,22 +306,36 @@ def _wrap_context(self, sock, sslopts, check_hostname=None, **ctx_options):
...
+        else:
+            # older versions of python 2.7 and python 2.6 do not have the
+            # ssl.PROTOCOL_TLS defined the equivalent is ssl.PROTOCOL_SSLv23
+            # we default to PROTOCOL_TLS and fallback to PROTOCOL_SSLv23
+            if hasattr(ssl, 'PROTOCOL_TLS'):
+                opts['ssl_version'] = ssl.PROTOCOL_TLS
+            else:
+                opts['ssl_version'] = ssl.PROTOCOL_SSLv23
+        # Set SNI headers if supported
+        if hasattr(ssl, 'HAS_SNI') and ssl.HAS_SNI:
```
georgepsarakis: I think you would probably want to check that `server_hostname` is also not `None` here? I guess this may trigger an error.

dhananjaysathe: In reality, None is an acceptable value for the SSLSocket. Pushed another commit anyway

georgepsarakis: Wasn't aware of this, thanks for sharing and fixing it either way :smiley: .

PR #153, opened 2017-06-12
Don't send AAAA DNS request when domain resolved to IPv4 address

booxter:
There is no need to send both A and AAAA requests when connecting to a
host since we are interested in a single address only. So if the host
resolves into IPv4 address, we can skip request for AAAA since it will
not be used anyway.

This sounds like a minor optimization, but it comes as a huge win in
case where DNS resolver is not configured for the requested host name,
but the name is listed in /etc/hosts with a IPv4 address. In this case,
resolution is very quick (the file is local, so no DNS request is sent),
except for the fact that we still ask for AAAA record for the host name.
A misconfigured DNS resolver can take time until it will time out the
request (30 seconds is a common length for Linux), which is an
unnecessary delay.

This exact situation hit OpenStack TripleO CI where DNS was not
configured, but resolution is implemented via /etc/hosts file which did
not include IPv6 entries.

OpenStack bug: https://bugs.launchpad.net/neutron/+bug/1696094

### DIFF
--- a/amqp/transport.py
+++ b/amqp/transport.py
@@ -151,9 +151,18 @@ class _AbstractTransport(object):
     def _connect(self, host, port, timeout):
         e = None
 
-        # avoid additional DNS requests for AAAA if A succeeds
+        # Below we are trying to avoid additional DNS requests for AAAA if A
+        # succeeds. This helps a lot in case when a hostname has an IPv4 entry
+        # in /etc/hosts but not IPv6. Without the (arguably somewhat twisted)
+        # logic below, getaddrinfo would attempt to resolve the hostname for
+        # both IP versions, which would make the resolver talk to configured
+        # DNS servers. If those servers are for some reason not available
+        # during resolution attempt (either because of system misconfiguration,
+        # or network connectivity problem), resolution process locks the
+        # _connect call for extended time.
         addr_types = (socket.AF_INET, socket.AF_INET6)
         for n, family in enumerate(addr_types):
+            # first, resolve the address for a single address family
             try:
                 entries = socket.getaddrinfo(
                     host, port, family, socket.SOCK_STREAM, SOL_TCP)
@@ -167,6 +176,8 @@ class _AbstractTransport(object):
                            if e is not None
                            else socket.error("failed to connect"))
                 continue
+
+            # now that we have address(es) for the hostname, connect to broker
             for i, res in enumerate(entries):
                 af, socktype, proto, canonname, sa = res
                 try:
@@ -185,6 +196,7 @@ class _AbstractTransport(object):
                     if i + 1 >= len(entries) and n + 1 >= len(addr_types):
                         raise
                 else:
+                    # hurray, we established connection
                     return
 
     def _init_socket(self, socket_settings, read_timeout, write_timeout):
--- a/amqp/transport.py
+++ b/amqp/transport.py
@@ -180,7 +180,7 @@ class _AbstractTransport(object):
 
             # now that we have address(es) for the hostname, connect to broker
             for i, res in enumerate(entries):
-                af, socktype, proto, canonname, sa = res
+                af, socktype, proto, _, sa = res
                 try:
                     self.sock = socket.socket(af, socktype, proto)
                     try:
--- a/amqp/transport.py
+++ b/amqp/transport.py
@@ -149,26 +149,43 @@ class _AbstractTransport(object):
                     sock.settimeout(prev)
 
     def _connect(self, host, port, timeout):
-        entries = socket.getaddrinfo(
-            host, port, 0, socket.SOCK_STREAM, SOL_TCP,
-        )
-        for i, res in enumerate(entries):
-            af, socktype, proto, canonname, sa = res
+        e = None
+
+        # avoid additional DNS requests for AAAA if A succeeds
+        addr_types = (socket.AF_INET, socket.AF_INET6)
+        for n, family in enumerate(addr_types):
             try:
-                self.sock = socket.socket(af, socktype, proto)
+                entries = socket.getaddrinfo(
+                    host, port, family, socket.SOCK_STREAM, SOL_TCP)
+            except socket.gaierror:
+                # we may have depleted all our options
+                if n + 1 >= len(addr_types):
+                    # if getaddrinfo succeeded before for another address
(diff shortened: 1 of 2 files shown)


booxter:
For my understanding, what's the expectation here regarding code coverage reduction? The _connect function is barely covered by any existing tests. Is it expected that I first write new tests for the code to then base my patch on top?

georgepsarakis:
@booxter a test that verifies the functionality is always beneficial, also for future readers. For this use case you can use `mock` to patch the calls to `socket.getaddrinfo` and `socket.socket`. I am guessing that coverage was slightly reduced because there are more branches in the flow, now that the loop was added.

booxter:
The failure in apicheck doesn't seem related to the patch.

georgepsarakis:
@booxter thanks for adding the test cases! I was wondering though, perhaps this is not a problem to be solved within the `py-amqp` library and can lead to other bugs / issues. IPv4/IPv6 preference can be controlled using the [/etc/gai.conf](https://linux.die.net/man/5/gai.conf) file, which controls the behavior of the `getaddrinfo` system call. `eventlet` that caused the [issue](https://github.com/eventlet/eventlet/commit/503aed1310b5e4a2fad9f5a0909663636a4e8752#diff-47bd8cdaac88cccf21624af4df7a35c3) you mention in the description, seems to be caused by custom DNS resolution code (which I am not sure why is needed in eventlet). Overall, I am not sure that this issue applied here.

booxter:
@georgepsarakis for what I understand, gai.conf does not control whether resolver fetches all records from all sources, but only the sorting order of the result. It means that with family=0, the work executed by the resolver will be the same irrespective of the file contents, even though the result may have different order of matching records. What I try to achieve here is to avoid some of the work for resolver if a satisfying record is found already. In this way, I can avoid a remote DNS call for AAAA in case when /etc/hosts contains a ipv4 address, which is totally fine for amqp.

An alternative could probably be making eventlet custom dns code to iterate through families one by one and yield results one by one if family=0 is passed, but even then we would need to introduce a change in amqp to short circuit the dns code on first successful record yielded from eventlet. Also it may be not compatible with stdlib getaddrinfo behavior that returns a list and not a generator.

amqp/transport.py
```diff
@@ -149,26 +149,43 @@ def having_timeout(self, timeout):
...
+                    host, port, family, socket.SOCK_STREAM, SOL_TCP)
+            except socket.gaierror:
+                # we may have depleted all our options
+                if n + 1 >= len(addr_types):
+                    # if getaddrinfo succeeded before for another address
+                    # family, reraise the previous socket.error since it's more
+                    # relevant to users
+                    raise (e
+                           if e is not None
+                           else socket.error("failed to connect"))
```
thedrow: Error message is not descriptive enough in my opinion. The message should at least mention that this is a DNS issue.

amqp/transport.py
```diff
@@ -149,26 +149,43 @@ def having_timeout(self, timeout):
...
+                    except NotImplementedError:
+                        pass
+                    self.sock.settimeout(timeout)
+                    self.sock.connect(sa)
+                except socket.error as e:
+                    if self.sock is not None:
+                        self.sock.close()
+                        self.sock = None
+                    # we may have depleted all our options
+                    if i + 1 >= len(entries) and n + 1 >= len(addr_types):
```
thedrow: While we're already touching this code let's extract the result of the `len()` calls to variables.

amqp/transport.py
```diff
@@ -149,26 +149,43 @@ def having_timeout(self, timeout):
...
 
     def _connect(self, host, port, timeout):
-        entries = socket.getaddrinfo(
-            host, port, 0, socket.SOCK_STREAM, SOL_TCP,
-        )
-        for i, res in enumerate(entries):
-            af, socktype, proto, canonname, sa = res
+        e = None
+
+        # avoid additional DNS requests for AAAA if A succeeds
```
thedrow: If we are going to complicate the logic here, at the very least let's leave a more meaningful comment that walks you through the process.

amqp/transport.py
```diff
@@ -149,26 +149,43 @@ def having_timeout(self, timeout):
...
+                # we may have depleted all our options
+                if n + 1 >= len(addr_types):
+                    # if getaddrinfo succeeded before for another address
+                    # family, reraise the previous socket.error since it's more
+                    # relevant to users
+                    raise (e
+                           if e is not None
+                           else socket.error("failed to connect"))
+                continue
+            for i, res in enumerate(entries):
```
thedrow: Add a newline after the end of the first for loop.

amqp/transport.py
```diff
@@ -149,26 +149,43 @@ def having_timeout(self, timeout):
...
+                af, socktype, proto, canonname, sa = res
```
thedrow: Since we don't use cannonname, turn it to _.

thedrow:
@booxter Will you be able to address my code review?

thedrow:
@booxter Ping?

amqp/transport.py
booxter: Gotcha. Added some comments in the method.

amqp/transport.py
booxter: Changed to "failed to resolve broker hostname"

amqp/transport.py
```diff
@@ -149,26 +149,58 @@ def having_timeout(self, timeout):
...
+                except socket.error as e:
+                    if self.sock is not None:
+                        self.sock.close()
+                        self.sock = None
+                    # we may have depleted all our options
+                    if i + 1 >= entries_num and n + 1 >= addr_types_num:
+                        raise
+                else:
+                    # hurray, we established connection
+                    return
```
thedrow: Tests do not cover this branch. If this is an easy fix I'll get to it myself

booxter: Do I read it right that you will take care of the unit test? Or do you expect it from me? (Trying not to do duplicate work.)

thedrow: I can get to it if you won't.

booxter: OK, please do since I won't be able to get to it in next days. Thanks a lot for your help and guidance!

thedrow:
Since you used master on your fork I can't push to your branch.

thedrow:
I'll just push the changes to master...

PR #154, opened 2017-07-07
publishing zero-length messages

### LINKED ISSUE #151: basic_publish_confirm hangs forever on empty message publish
## Scenario ##

You have an empty message you'd like to publish for whatever reason. (We post a serialized protobuf on a queue, and if that protobuf is nothing but the default proto values, it ends up being `b''`.

You also use publisher confirms.

## What happens ##

Calling `basic_publish_confirm` with a message that has a body of `b''` does send some headers over the wire, but rabbitmq never thinks the message is complete. If you then publish another message, you see an error in the rabbitmq logs:

```
=ERROR REPORT==== 31-May-2017::20:12:42 ===
Error on AMQP connection <0.2299.1> (172.17.0.1:55824 -> 172.17.0.4:5672, vhost: '/', user: 'guest', state: running), channel 1:
operation basic.publish caused a connection exception unexpected_frame: "expected content header for class 60, got non content header frame instead"
```

## Debug ##

I traced this all the way down to where it sets the network packet up and packages up the struct and such, I'm a bit out of my depth at this point, but my guess is that there needs to be some sort of "end of body" token and that's not getting set. I definitely see it send the headers, skipping over the body (as it's empty).

If you call `basic_publish` (no confirm) the call returns immediately, but nothing is ever truly published, still an error when the next message comes across in rabbit.

## Other libraries where this works ##

I only tested one, but [Pika](https://pika.readthedocs.io/en/0.10.0/) does publish and confirm the empty message successfully.

## Example Code ##

```
In [1]: from amqp import Connection, Message
In [2]: con = Connection(host='localhost', userid='guest', password='guest', virtual_host='/')
In [3]: con.connect()
In [4]: chan = con.channel()
In [5]: chan.basic_publish_confirm(Message(b''), 'system-configuration', 'a.config.c')
# hangs forever here
```

## Versions ##

[amqp 2.1.4](https://pypi.python.org/pypi/amqp/2.1.4)

[RabbitMQ 3.6.6](https://hub.docker.com/_/rabbitmq/)

Let me know if I can provide any other details.
Thanks

christian-blades-cb:
RabbitMQ expects a header frame and no body frame when the body is zero length. This may also be true of other 0.9.x AMQP servers, but I haven't tested those.

fixes #151

### DIFF
--- a/amqp/method_framing.py
+++ b/amqp/method_framing.py
@@ -159,7 +159,7 @@ def frame_writer(connection, transport,
             pack_into('>BHI%dsB' % framelen, buf, offset,
                       type_, channel, framelen, frame, 0xce)
             offset += 8 + framelen
-            if body:
+            if body is not None:
                 frame = b''.join([
                     pack('>HHQ', method_sig[0], 0, len(body)),
                     properties,
@@ -170,10 +170,12 @@ def frame_writer(connection, transport,
                           2, channel, framelen, frame, 0xce)
                 offset += 8 + framelen
 
-                framelen = len(body)
-                pack_into('>BHI%dsB' % framelen, buf, offset,
-                          3, channel, framelen, str_to_bytes(body), 0xce)
-                offset += 8 + framelen
+                bodylen = len(body)
+                if bodylen > 0:
+                    framelen = bodylen
+                    pack_into('>BHI%dsB' % framelen, buf, offset,
+                              3, channel, framelen, str_to_bytes(body), 0xce)
+                    offset += 8 + framelen
 
             write(view[:offset])
 
--- a/t/unit/test_method_framing.py
+++ b/t/unit/test_method_framing.py
@@ -99,3 +99,9 @@ class test_frame_writer:
         frame = 2, 1, spec.Basic.Publish, b'x' * 10, msg
         self.g(*frame)
         self.write.assert_called()
+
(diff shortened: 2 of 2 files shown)


christian-blades-cb:
I'm gonna need some help interpreting that failure in `apicheck`.

auvipy:
is it ready for review?

christian-blades-cb:
@auvipy yes, please

christian-blades-cb:
@auvipy any word on reviews?

Changelog
```diff
@@ -5,7 +5,20 @@ py-amqp is fork of amqplib used by Kombu containing additional features and impr
 The previous amqplib changelog is here:
 http://code.google.com/p/py-amqplib/source/browse/CHANGES
 
-.. _version-2.2.0
+.. _version-2.2.1:
```
thedrow: This change is not related to this PR. Can you please rebase?

christian-blades-cb: @thedrow: I removed your commit, but now:
```
This branch has conflicts that must be resolved
Only those with write access to this repository can merge pull requests.
Conflicting files
Changelog
```

I vaguely recall that this is why that commit was in this PR to begin with: rebasing to resolve the merge conflict.

christian-blades-cb: Oh, I see now, the colon in particular. Got it.

Issue #82, opened 2016-04-22
Setup.py PyPi licensing metadata and LICENSE file are not aligned


gwachob:
Hi folks-
   I'm tasked with going through the licensing on  dependencies on a project, and I've noticed that the licensing trove classifiers indicate both BSD and LGPL, but the LICENSE file only contains LGPL. Can this be clarified (either by removing the line from the setup.py or modifying/extending the LICENSE file)?

ask:
Ooops, I think the license needs to stay LGPL unless we can get permission from Barry Pederson.

This is originally a fork of his amqplib project which is LGPL, though the code is almost entirely rewritten at this point.

cwill747:
+1 on this - weird to me that everything I've seen in the celery project is not licensed LGPL but this is - would be nice to align to one license

cwill747:
@barryp thoughts on BSD 3-clause-ing this code?

barryp:
I don't have strong feelings about LGPL vs BSD, so if switching your fork helps you guys out I say go for it.

ask:
@barryp Thank you so much for amqplib and other contributions, there wouldn't be a Celery without you!  I'm very happy we can unify the license :tada:

barryp:
@ask You're very welcome!  I'm thrilled that it's being used in such a cool project.

PR #163, opened 2017-08-22
Remove LGPL headers from source files

FatalEnigma:
Project has since changed to BSD


thedrow:
Hmm I was about to hit the merge button but I saw that remark about the fork from amqplib.
I think @barryp owns a tiny bit of the code here (since it has changed so much since the fork) so we'd have to get his permission first to remove these headers.

@barryp Hi Barry :)
Do you mind if we'll relicense these files under BSD?

barryp:
BSD is fine, this was covered back in issue #82

thedrow:
Thanks @barryp :)

PR #164, opened 2017-08-30
Update UTF-8 encoding behavior between Py2/Py3

tylerjharden:
Python 2 defaults to 'surrogatepass' behavior when encoding a string with invalid UTF-8 bytes, in that it will allow surrogates to passthrough. Python 3 throws errors by default. This means systems using Celery with AMQP that deal with passing around potentially bunk content will fail out.

I made a similar change to core ElasticSearch Python libraries to solve similar issues there.

Resolves https://github.com/celery/py-amqp/issues/165

### DIFF
--- a/amqp/serialization.py
+++ b/amqp/serialization.py
@@ -290,14 +290,14 @@ def dumps(format, values):
             val = val or ''
             bitcount = _flushbits(bits, write)
             if isinstance(val, string):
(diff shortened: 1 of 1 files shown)


tylerjharden:
I'd like to note that the CI failure is not relevant to my changes, something with your configuration.

PR #170, opened 2017-10-17
Fix ResourceWarning in setup.py

jdufresne:
When Python is executed with warnings enabled (e.g. `python3 -Wall setup.py ...`), `ResourceWarning`s appear on stderr. This occurs because files are not explicitly closed. Always use a context manager to close files when finished.


thedrow:
Thanks!

PR #176, opened 2017-12-22
Support for EXTERNAL authentication and specific login_method

merll:
This PR addresses two issues that can occur when using the library as part of Celery 4.x in combination with RabbitMQ:

1. The EXTERNAL authentication is not supported. This could also be added by implementing it in our application and passing it through the `authentication` parameter; but since it is mentioned in the [Kombu docs](http://docs.celeryproject.org/projects/kombu/en/latest/reference/kombu.connection.html), I think it should also be implemented directly in `py-amqp`.

2. Setting the `login_method` parameter, as suggested in the [Celery configuration docs](http://docs.celeryproject.org/en/latest/userguide/configuration.html#broker-login-method) or the [Kombu docs](http://docs.celeryproject.org/projects/kombu/en/latest/reference/kombu.connection.html) does not work as expected. It is not possible to e.g. force using `GSSAPI` instead of `AMQPLAIN` and fail if the server does not support it. The only way to do this is using the `authentication` parameter, passing in the `sasl` object.

These commits add an `EXTERNAL` authentication class to the `sasl` module, and to the list of mechanisms supported where demanded by the server. In addition, the  `login_method` parameter forces a specific method, also considering that `EXTERNAL` and `GSSAPI` do not require `userid` or `password` to be set.

### DIFF
--- a/amqp/connection.py
+++ b/amqp/connection.py
@@ -190,6 +190,7 @@ class Connection(AbstractChannel):
             self.authentication = (sasl.RAW(login_method, login_response),)
         elif userid is not None and password is not None:
             self.authentication = (sasl.GSSAPI(userid, fail_soft=True),
+                                   sasl.EXTERNAL(),
                                    sasl.AMQPLAIN(userid, password),
                                    sasl.PLAIN(userid, password))
         else:
--- a/amqp/connection.py
+++ b/amqp/connection.py
@@ -87,7 +87,7 @@ class Connection(AbstractChannel):
     'GSSAPI', 'EXTERNAL', 'AMQPLAIN', or 'PLAIN'.
     Otherwise authentication will be performed using any supported method
     preferred by the server. Userid and passwords apply to AMQPLAIN and
-    PLAIN authentication, whereas on GSSAPI only userid will be used as
+    PLAIN authentication, whereas on GSSAPI only userid will be used as the
     client name. For EXTERNAL authentication both userid and password are
     ignored.
 
--- a/amqp/connection.py
+++ b/amqp/connection.py
@@ -199,11 +199,13 @@ class Connection(AbstractChannel):
                 auth = sasl.EXTERNAL()
             elif login_method == 'AMQPLAIN':
                 if userid is None or password is None:
-                    raise ValueError("Must supply authentication or userid/password")
+                    raise ValueError(
+                        "Must supply authentication or userid/password")
                 auth = sasl.AMQPLAIN(userid, password)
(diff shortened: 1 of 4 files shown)


amqp/connection.py
```diff
@@ -83,7 +83,13 @@ class Connection(AbstractChannel):
 
     Authentication can be controlled by passing one or more
     `amqp.sasl.SASL` instances as the `authentication` parameter, or
-    by using the userid and password parameters (for AMQPLAIN and PLAIN).
+    setting the `login_method` string to one of the supported methods:
+    'GSSAPI', 'EXTERNAL', 'AMQPLAIN', or 'PLAIN'.
+    Otherwise authentication will be performed using any supported method
+    preferred by the server. Userid and passwords apply to AMQPLAIN and
+    PLAIN authentication, whereas on GSSAPI only userid will be used as
+    client name. For EXTERNAL authentication both userid and password are
```
thedrow: Please adjust comment to "whereas on GSSAPI only userid will be used as **the** client name"

amqp/sasl.py
```diff
@@ -36,6 +36,8 @@ def __init__(self, username, password):
         self.username, self.password = username, password
 
     def start(self, connection):
+        if self.username is None or self.password is None:
+            return NotImplemented
```
thedrow: Why are we returning NotImplemented? It doesn't seem like it's the right thing to do.
See https://stackoverflow.com/a/44575926/85140

amqp/sasl.py
```diff
@@ -56,6 +58,8 @@ def __init__(self, username, password):
...
```
thedrow: Same as the comment above.

amqp/sasl.py
merll: I had reviewed #110, more specifically the comment of 5f63857f6af350309652bd2ee19a3b6531af33eb, and thought I should adjust to that. However I do agree it does not seem right. What alternatives would you prefer?

One option would be, if credentials are missing, not to add them to `self.authentication` in the first place. In that case I would prefer moving the argument availability checks to the `SASL` classes entirely so that the `__init__()` method in `Connection` does not get too complex.

The second option would simply be to return something else than `NotImplemented`, e.g. `None` or a custom exception during `start()`.

thedrow: If we were to change that, I'd do so in a different PR. I just didn't know that's the way it currently works.

thedrow:
Thanks for the contribution! Feel free to add yourself to the AUTHORS file.

PR #187, opened 2018-03-22
for deserialisation, correct offset for bits

allan-simon:
it needs to be increased only at the beginning and once every 8 bits

otherwise typically it was failling to decode "queue.declare"

### DIFF
--- a/amqp/serialization.py
+++ b/amqp/serialization.py
@@ -165,11 +165,12 @@ def loads(format, buf, offset=0,
         if p == 'b':
             if not bitcount:
                 bits = ord(buf[offset:offset + 1])
+                offset += 1
             bitcount = 8
             val = (bits & 1) == 1
             bits >>= 1
             bitcount -= 1
-            offset += 1
+
         elif p == 'o':
             bitcount = bits = 0
             val, = unpack_from('>B', buf, offset)
--- a/t/unit/test_serialization.py
+++ b/t/unit/test_serialization.py
@@ -97,7 +97,8 @@ class test_serialization:
             dumps('A', [[object()]])
 
     def test_bit_offset_adjusted_correctly(self):
-        buf = dumps('BssbbbbbF', [50, "quick", "fox", True, False, False, True, True, {"prop1": True}])
+        buf = dumps('BssbbbbbF', [50, "quick", "fox", True,
+                                  False, False, True, True, {"prop1": True}])
         loads('BssbbbbbF', buf)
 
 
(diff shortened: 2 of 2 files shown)


thedrow:
Has no one been bitten by this bug before? This code was there for a long time.

Can you explain in more detail what are you trying to fix?

allan-simon:
I think it's because this code is only used when you implement a AMQP server? (i found it when implementing the exchange-declare method which is one of the few to have "packed bits" )

I don't think I will have time to dig into how you implement tests and write a regression case.

thedrow:
We can't merge a PR without making sure no regressions has been introduced.
The build passes so it gives us some confidence that this change doesn't break anything but we need tests to merge this.

If you can, at the very least, provide us with a way to reproduce the issue somehow we'll write the tests for it.

thedrow:
That's a very cool project.
We can probably use it ourselves.
I'll try to dig in and come up with a test case.

thedrow:
I finally managed to find out how to write a test for this.
If you'll try to run the test without this patch, the test fails.

allan-simon:
nice thanks for taking the time to write the test case :)

thedrow:
While we're at it, can you please explain why we set the bitset to 8 and than decrease it by 1?
Most types reset it anyway.
I'm not sure I understand the algorithm and I think it requires some inline comments to explain it.

allan-simon:
it has been a year, but if i'm not mistaken  it's because if you have a consecutive bit values you continue to "consume" the same byte as  the next bit  will be store in the same byte (so you can pack up to 8 bit values this way, after which you need to proceed to the next byte anyway)

thedrow:
And that doesn't happen in frames or arrays apparently. Interesting. Thanks.

thedrow:
So why aren't we reseting the bitcount when we encounter a byte array? Got a clue?

allan-simon:
@thedrow  which lines are you referring ?

thedrow:
@allan-simon https://github.com/celery/py-amqp/blob/master/amqp/serialization.py#L213

thedrow:
It turns out the code should be:
```
if not bitcount:
                bits = ord(buf[offset:offset + 1])
                offset += 1
                bitcount = 8
val = (bits & 1) == 1
bits >>= 1
bitcount -= 1
```

Because when we have 16 boolean flags packed into 8bit bitsets we fail to parse them correctly.
See #260 for details.

You might want to update your code as well or adjust it to reuse ours.

allan-simon:
nice thanks for the precision I will update it :)

PR #202, opened 2018-08-08
AMQP protocol header for AMQP 0-9-1

carlhoerberg:
See https://www.rabbitmq.com/resources/specs/amqp0-9-1.pdf chapter 4.2.2


thedrow:
I wonder how this was wrong all this time...

PR #210, opened 2018-10-09
Issue#209: Connection.connect, Connection.close don't close socket on exception

tomc797:
This PR resolves the problem discussed in Issue#209. In summary, the problem is that an exception during Connection.open, Connection.close leaves the socket open. The PR addresses the problem by appropriately closing the socket when an exception is encountered.

### DIFF
--- a/amqp/connection.py
+++ b/amqp/connection.py
@@ -294,18 +294,24 @@ class Connection(AbstractChannel):
         #
         if self.connected:
             return callback() if callback else None
-        self.transport = self.Transport(
-            self.host, self.connect_timeout, self.ssl,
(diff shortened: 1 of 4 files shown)


tomc797:
Good day auvipy -

I've added the tests necessary to exercise the code changes.

There is a change to the Connect.connect behavior. Before, a timeout could interrupt the handshake_complete flag check and the connection would remain in a valid state and the handshake would eventually complete if everything is successfully. With the changes in this PR, the handshake_complete must be set for success; otherwise, the underlying transport is closed. The reason I developed the code this way is that a user cannot distinguish between a failed connect attempt and a connect attempt that may eventually succeed.

Issue #200, opened 2018-07-17
Openbsd 6.3 error: Protocol not available


raratiru:
* Openbsd-6.3
* Python-3.6
* Celery-4.2.0
* amqp-2.3.2

When starting Celery, I receive the error:

`[2018-07-18 00:53:50,510: ERROR/MainProcess] consumer: Cannot connect to amqp://...: [Errno 42] Protocol not available.`

**amqp-2.2.2** is the most recent version that works.

neostimul:
The same situation with FreeBSD 11.2-RELEASE

auvipy:
plz send a pr to fix the regression

jasconius:
confirming this issue on FreeBSD. Using freebsd's package manager version of rabbitmq, but identical behavior... going to attempt to roll back to older rabbitmq to see if this addresses it

clokep:
You likely need to modify the `KNOWN_TCP_OPTS` on FreeBSD, it seems to already have a bunch of platform specific changes in it.

raratiru:
> You likely need to modify the `KNOWN_TCP_OPTS` on FreeBSD, it seems to already have a bunch of platform specific changes in it.

@clokep  [I just got confused! ](https://www.google.com/search?client=firefox-b&ei=W7O7W4ykIbGk_QaGwa6ADg&q=KNOWN_TCP_OPTS+freebsd&oq=KNOWN_TCP_OPTS+freebsd&gs_l=psy-ab.3...0.0.0.3176.0.0.0.0.0.0.0.0..0.0....0...1c..64.psy-ab..0.0.0....0.WnvE8AmJZCs) What kind of modification is requied?

clokep:
@raratiru Check out the changes in #167, they touch this code that configures `KNOWN_TCP_OPTS`:

https://github.com/celery/py-amqp/blob/dc5b07c2320f932b2dd8342d0f4d578c095600cf/amqp/platform.py#L32-L57

Something is probably "wrong" with this code for FreeBSD. You could check what this code returns and attempt removing each option until you don't get this error. (Or compare what `KNOWN_TCP_OPTS` ends up as between 2.2.2 and the latest version.)

raratiru:
OK, this is clear!

The author of #167 is referring to some failing tests however that [are unreachable now](https://codecov.io/gh/celery/py-amqp/pull/167?src=pr&el=tree).

> I've added some tests around this. Once again - I don't think I'm responsible for failing tests.

[...]

PR #211, opened 2018-10-12
Fix Protocol not available for BSD (#200)

raratiru:
TCP_USER_TIMEOUT has to be excluded from KNOWN_TCP_OPTS in bsd platforms

### DIFF
--- a/amqp/platform.py
+++ b/amqp/platform.py
@@ -51,6 +51,9 @@ if sys.platform.startswith('linux'):
 elif sys.platform.startswith('darwin'):
     KNOWN_TCP_OPTS.remove('TCP_USER_TIMEOUT')
 
+elif 'bsd' in sys.platform:
+    KNOWN_TCP_OPTS.remove('TCP_USER_TIMEOUT')
+
 # According to MSDN Windows platforms support getsockopt(TCP_MAXSSEG) but not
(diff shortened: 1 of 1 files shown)


raratiru:
According to [this answer](https://stackoverflow.com/a/13874620/2996101), `sys.platform` gives the following results:

> 
>     .---------------------.------------.
>     | System              | Value      |
>     |---------------------|------------|
>     | Linux (2.x and 3.x) | linux2 (*) |
>     | Windows             | win32      |
>     | Windows/Cygwin      | cygwin     |
>     | Mac OS X            | darwin     |
>     | OS/2                | os2        |
>     | OS/2 EMX            | os2emx     |
>     | RiscOS              | riscos     |
>     | AtheOS              | atheos     |
>     | FreeBSD 7           | freebsd7   |
>     | FreeBSD 8           | freebsd8   |
>     | FreeBSD N           | freebsdN   |
>     | OpenBSD 6           | openbsd6   |
>     '---------------------'------------'
> (*) Starting from python3.3, the value for any Linux is always linux (no version number).

Therefore the string `'bsd'` could be a catch-all solution.

clokep:
@auvipy I Don't think it's possible unless there's CI on *BSD. This feels reasonable to me though since macOS / Darwin is BSD-y under the hood.

Issue #206, opened 2018-09-19
Doesn't work with Python 3.7


galuszkak:
This is stack trace in Python 3.7 when using amqp with SSL RabbitMQ. 
@thedrow @auvipy @ask 

Sounds it's kinda critical to move to python 3.7
```
  File "/usr/local/lib/python3.7/site-packages/amqp/connection.py", line 302, in connect
    self.transport.connect()
  File "/usr/local/lib/python3.7/site-packages/amqp/transport.py", line 81, in connect
    self.socket_settings, self.read_timeout, self.write_timeout,
  File "/usr/local/lib/python3.7/site-packages/amqp/transport.py", line 176, in _init_socket
    self._setup_transport()
  File "/usr/local/lib/python3.7/site-packages/amqp/transport.py", line 295, in _setup_transport
    self.sock = self._wrap_socket(self.sock, **self.sslopts)
  File "/usr/local/lib/python3.7/site-packages/amqp/transport.py", line 302, in _wrap_socket
    return self._wrap_socket_sni(sock, **sslopts)
  File "/usr/local/lib/python3.7/site-packages/amqp/transport.py", line 341, in _wrap_socket_sni
    sock = ssl.SSLSocket(**opts)
  File "/usr/local/lib/python3.7/ssl.py", line 788, in __init__
    f"{self.__class__.__name__} does not have a public "
```

auvipy:
thanks for the report, are you trying to use the master? or the latest pypi release which is not yet compatible?

galuszkak:
@auvipy both doesn't work.

auvipy:
I will try it locally now and possibly fix too

auvipy:
btw are you using eventlet? https://github.com/eventlet/eventlet/pull/506

thedrow:
Is this the full stack trace?
That doesn't look complete.

galuszkak:
@thedrow I'm not sure there is much more to add. Here is full stack trace:
```
  File "/app/workflows/rabbitmq/manager.py", line 24, in connect
    self.connection.connect()
  File "/usr/local/lib/python3.7/site-packages/kombu/connection.py", line 261, in connect
    return self.connection
  File "/usr/local/lib/python3.7/site-packages/kombu/connection.py", line 802, in connection
    self._connection = self._establish_connection()
  File "/usr/local/lib/python3.7/site-packages/kombu/connection.py", line 757, in _establish_connection
    conn = self.transport.establish_connection()
[...]

PR #215, opened 2018-11-05
Change direct SSLSocket instantiation with wrap_socket

avborhanian:
Attempt at fixing the 3.7 issue.

### DIFF
--- a/amqp/transport.py
+++ b/amqp/transport.py
@@ -341,11 +341,16 @@ class SSLTransport(_AbstractTransport):
                 opts['ssl_version'] = ssl.PROTOCOL_TLS
             else:
                 opts['ssl_version'] = ssl.PROTOCOL_SSLv23
+        sock = ssl.wrap_socket(**opts)
         # Set SNI headers if supported
         if (server_hostname is not None) and (
-                hasattr(ssl, 'HAS_SNI') and ssl.HAS_SNI):
-            opts['server_hostname'] = server_hostname
-        sock = ssl.SSLSocket(**opts)
+                hasattr(ssl, 'HAS_SNI') and ssl.HAS_SNI) and (
+                hasattr(ssl, 'SSLContext')):
+            context = ssl.SSLContext(opts['ssl_version'])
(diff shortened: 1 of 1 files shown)


clokep:
Related to #206.

clokep:
I tested this briefly in my testing environment which uses amqps via py-amqp on Python 2.7.6 on Ubuntu 14.04. I was able to connect OK to RabbitMQ with this change.

thedrow:
I'll release this as soon as I can.

auvipy:
If you document the release process, I can release more frequently hopefully

dhananjaysathe:
This breaks on python 2.7 the wrap_context method returns out for some reason and does not execute any further code the wrapping does not happen and you end up with an invalid sni header. Do we have a test that actually tests for SNI ?

amqp/transport.py
```diff
@@ -341,11 +341,16 @@ def _wrap_socket_sni(self, sock, keyfile=None, certfile=None,
...
         # Set SNI headers if supported
         if (server_hostname is not None) and (
-                hasattr(ssl, 'HAS_SNI') and ssl.HAS_SNI):
-            opts['server_hostname'] = server_hostname
-        sock = ssl.SSLSocket(**opts)
+                hasattr(ssl, 'HAS_SNI') and ssl.HAS_SNI) and (
+                hasattr(ssl, 'SSLContext')):
+            context = ssl.SSLContext(opts['ssl_version'])
+            context.verify_mode = cert_reqs
+            context.check_hostname = True
```
dhananjaysathe: why is this line being set here ?

amqp/transport.py
```diff
@@ -341,11 +341,16 @@ def _wrap_socket_sni(self, sock, keyfile=None, certfile=None,
...
+            context.load_cert_chain(certfile, keyfile)
```
dhananjaysathe: also if this was none this should not be executed

thedrow: Feel free to send a PR.

PR #213, opened 2018-10-26
Added integration tests

matusvalo:
This PR adds integration tests from py-amqp library. This tests verifies that proper callback is called when given method is received from network. This tests injects data at the lowest level possible in the `Transport.read_frame()` method. Integration of `Connection`, `Channel` and `AbstractChannel` classes is tested. Also the logic can be reused/extended to test more functionality of the library

PR #216, opened 2018-11-06
Prevent failed checks in integration tests caused by table datatype

matusvalo:
This PR fixes issues in integration test caused by random order of items in Table amqp type. Instead of direct comparison of serialized byte stream we compare deserialized data structures.


matusvalo:
This PR fixes issue in PR #213.

Issue #218, opened 2018-11-13
When py-amqp tries to publish to nonexistent exchange crash with AttributeError: 'NoneType' object has no attribute 'drain_events'


matusvalo:
Steps to reproduce:
1. ensure that `test_exc` exchange does not exist
2. execute the code below

Example code:
```python
import amqp

conn = amqp.Connection(host='localhost')
conn.connect()
ch = conn.channel(channel_id=1)
ch.basic_publish(msg=amqp.Message(body=b'Hello world'), exchange='test_exc', routing_key='rkey')
conn.close()
```
Executing of the script ends up with the following stacktrace:
```Traceback (most recent call last):
  File "test.py", line 7, in <module>
    conn.close()
  File "/home/matus/dev/py-amqp/amqp/connection.py", line 575, in close
    wait=spec.Connection.CloseOk,
  File "/home/matus/dev/py-amqp/amqp/abstract_channel.py", line 59, in send_method
    return self.wait(wait, returns_tuple=returns_tuple)
  File "/home/matus/dev/py-amqp/amqp/abstract_channel.py", line 79, in wait
    self.connection.drain_events(timeout=timeout)
  File "/home/matus/dev/py-amqp/amqp/connection.py", line 500, in drain_events
    while not self.blocking_read(timeout):
  File "/home/matus/dev/py-amqp/amqp/connection.py", line 506, in blocking_read
    return self.on_inbound_frame(frame)
  File "/home/matus/dev/py-amqp/amqp/method_framing.py", line 55, in on_frame
    callback(channel, method_sig, buf, None)
  File "/home/matus/dev/py-amqp/amqp/connection.py", line 510, in on_inbound_method
    method_sig, payload, content,
  File "/home/matus/dev/py-amqp/amqp/abstract_channel.py", line 124, in dispatch_method
    listener(*args)
  File "/home/matus/dev/py-amqp/amqp/channel.py", line 277, in _on_close
    self._do_revive()
  File "/home/matus/dev/py-amqp/amqp/channel.py", line 164, in _do_revive
    self.open()
  File "/home/matus/dev/py-amqp/amqp/channel.py", line 434, in open
    spec.Channel.Open, 's', ('',), wait=spec.Channel.OpenOk,
  File "/home/matus/dev/py-amqp/amqp/abstract_channel.py", line 59, in send_method
    return self.wait(wait, returns_tuple=returns_tuple)
[...]

PR #220, opened 2018-11-14
Don't revive channel when connection is closing

matusvalo:
This is POC fixing #218. 
BEFORE:
```
]# python test.py
  File "test.py", line 7, in <module>
    conn.close()
  File "/home/matus/dev/py-amqp/amqp/connection.py", line 578, in close
    wait=spec.Connection.CloseOk,
  File "/home/matus/dev/py-amqp/amqp/abstract_channel.py", line 61, in send_method
    return self.wait(wait, returns_tuple=returns_tuple)
  File "/home/matus/dev/py-amqp/amqp/abstract_channel.py", line 84, in wait
    self.connection.drain_events(timeout=timeout)
  File "/home/matus/dev/py-amqp/amqp/connection.py", line 502, in drain_events
    while not self.blocking_read(timeout):
  File "/home/matus/dev/py-amqp/amqp/connection.py", line 508, in blocking_read
    return self.on_inbound_frame(frame)
  File "/home/matus/dev/py-amqp/amqp/method_framing.py", line 55, in on_frame
    callback(channel, method_sig, buf, None)
  File "/home/matus/dev/py-amqp/amqp/connection.py", line 512, in on_inbound_method
    method_sig, payload, content,
  File "/home/matus/dev/py-amqp/amqp/abstract_channel.py", line 129, in dispatch_method
    listener(*args)
  File "/home/matus/dev/py-amqp/amqp/channel.py", line 280, in _on_close
    self._do_revive()
  File "/home/matus/dev/py-amqp/amqp/channel.py", line 163, in _do_revive
    self.open()
  File "/home/matus/dev/py-amqp/amqp/channel.py", line 437, in open
    spec.Channel.Open, 's', ('',), wait=spec.Channel.OpenOk,
  File "/home/matus/dev/py-amqp/amqp/abstract_channel.py", line 61, in send_method
    return self.wait(wait, returns_tuple=returns_tuple)
  File "/home/matus/dev/py-amqp/amqp/abstract_channel.py", line 84, in wait
    self.connection.drain_events(timeout=timeout)
AttributeError: 'NoneType' object has no attribute 'drain_events'
```
AFTER:
```
# python test.py
  File "test.py", line 7, in <module>
    conn.close()
  File "/home/matus/dev/py-amqp/amqp/connection.py", line 578, in close
    wait=spec.Connection.CloseOk,
  File "/home/matus/dev/py-amqp/amqp/abstract_channel.py", line 61, in send_method
    return self.wait(wait, returns_tuple=returns_tuple)
  File "/home/matus/dev/py-amqp/amqp/abstract_channel.py", line 84, in wait
    self.connection.drain_events(timeout=timeout)
  File "/home/matus/dev/py-amqp/amqp/connection.py", line 502, in drain_events
    while not self.blocking_read(timeout):
  File "/home/matus/dev/py-amqp/amqp/connection.py", line 508, in blocking_read
    return self.on_inbound_frame(frame)
  File "/home/matus/dev/py-amqp/amqp/method_framing.py", line 55, in on_frame
    callback(channel, method_sig, buf, None)
  File "/home/matus/dev/py-amqp/amqp/connection.py", line 512, in on_inbound_method
    method_sig, payload, content,
  File "/home/matus/dev/py-amqp/amqp/abstract_channel.py", line 129, in dispatch_method
    listener(*args)
  File "/home/matus/dev/py-amqp/amqp/channel.py", line 281, in _on_close
    reply_code, reply_text, (class_id, method_id), ChannelError,
amqp.exceptions.NotFound: Basic.publish: (404) NOT_FOUND - no exchange 'test_exc' in vhost '/'
```
Unittests are for now failing. Is this right way to fix issue #218 ? Honestly I don't like adding new state to Channel/Connection. Is there any rationale/usage of reviving channel after close? Is it used in kombu or Celery? Or when closing channel/connection all messages except CloseOk should be silently ignored?

The reviving of channel was introduced in this commit: https://github.com/celery/py-amqp/commit/34f47580b09a7a48fbbec25f0017646114b26c36


thedrow:
I don't really know what @ask did or why. If there's a reference to an issue, it'd be nice to read the reasoning behind it.
You should adjust the unit test since now channels do not always revive.
To me this looks like the right fix.
@georgepsarakis @auvipy What do you guys think?

auvipy:
I'm OK with it. is it possible to test this? if not we can include it and release 2.4 version

amqp/channel.py
```diff
@@ -219,12 +219,14 @@ def close(self, reply_code=0, reply_text='', method_sig=(0, 0),
             if is_closed:
                 return
 
+            self.is_closing = True
```
georgepsarakis: This property does not seem to be used. Perhaps you considered it as an initial implementation?

amqp/connection.py
```diff
@@ -569,12 +569,14 @@ def close(self, reply_code=0, reply_text='', method_sig=(0, 0),
             return
 
         try:
+            self.is_closing = True
```
georgepsarakis: What happens if you call [Connection.collect](https://github.com/celery/py-amqp/pull/220/files#diff-2a6a0da3750a5417c3a6341ad912f551R452) here, instead of handling this when `Channel._do_revive` is called?

amqp/channel.py
matusvalo: I decided do the state more generic to cover also channels not only connection (since connection is basically special channel with id 0).

amqp/connection.py
matusvalo: It cannot be done. Client must wait for CloseOk method and after do the clean up - see the close method docs [1]

[1] https://www.rabbitmq.com/amqp-0-9-1-reference.html

matusvalo:
@georgepsarakis @thedrow  @auvipy tests can be done using integration tests. They are able to simulate pretty any method exchange with mocked broker.

What is needed to be done and can bite later is that when sending back `CloseOk` using `AbstractChannel.wait()` which is internally calling `Connection.drain_events()`.  The problem is that `drain_events()` method reacts to **all** methods coming from server. This is not allowed from docs [1]:

> close(reply-code reply-code, reply-text reply-text, class-id class-id, method-id method-id) ➔ close-ok
> Request a connection close.
> 
> This method indicates that the sender wants to close the connection. This may be due to internal conditions (e.g. a forced shut-down) or due to an error handling a specific method, i.e. an exception. When a close is due to an exception, the sender provides the class and method id of the method which caused the exception.
> 
> After sending this method, any received methods except Close and Close-OK MUST be discarded. The response to receiving a Close after sending Close must be to send Close-Ok. 

Hence, if you are fine with proposed fix, I will create tests covering this functionality. After I will create new issue touching the problem with `drain_events()` method.

[1] https://www.rabbitmq.com/amqp-0-9-1-reference.html

matusvalo:
I have added test covering this situation - see `test_received_channel_Close_during_connection_close()`
Basically, I am simulating the following situation:
1. User requests closing connection - client sends Connection.Close
2. Broker requests closing Channel - client receives Channel.Close
3. Broker sends Connection.CloseOk

This test is broken when fix of PR is not applied.

Btw. I have tried recently to reproduce Issue #218 with real broker but I was not able to. I am not sure why it happened before. Maybe some specific timing situation on broker side? Maybe I did something wrong. I am not sure now. Hence, I am not sure whether this PR should be merged or not... :-(

PR #208, opened 2018-10-03
Handle negative acknowledgments sent by RMQ

matusvalo:
This PR fixes #207. This PR:
* adds nack specs and callbacks to channel
* adds nack method and callback in `channel.wait()`
* `channel.wait()` callbacks now has AMQP method as first parameter
* `channel.basic_publish_confirm()` now raises `MessageNacked` exception when library receives Nack when waiting for confirmation

PR #233, opened 2018-12-31
Don't return method signature from AbstractChannel::wait()

### LINKED ISSUE #232: channel.queue_declare raises TypeError
PR #208 Introduces calls callbacks of wait with new parameter (`method`). This new parameter is returned via `wait()` method which breaks some code which does not expect this new parameter to be returned by `wait()` method.

Example broken script:

```python
import amqp
conn = amqp.Connection()
conn.connect()
ch = conn.channel()
ch.queue_declare('foo')
```
This script crash with exception:
```
TypeError: __new__() takes 4 positional arguments but 5 were given
```
The easy hot fix is as follows in `AbstractChannel::wait`
```python
            if p.value:
                args, kwargs = p.value
                args = args[1:]   # don`t return method parameter
                return args if returns_tuple else (args and args[0])
```

matusvalo:
PR #208 added new parameter `method_sig` for `wait()` callback. This change breaks some parts of existing code. This commit adjust `wait()` method to not return `method_sig` and hence to be retro compatible with the rest of the library.

Closes #232


matusvalo:
PR also introduces test `test_queue_declare` which checks declaring of queue. This was broken due #232

PR #235, opened 2019-01-10
Move delivery_info to constructor of Message.

matusvalo:
This PR fixes delivery_info attribute of Message class which is defined as static attribute but is used as instance attribute. This PR also cleans and improves tests.


auvipy:
any related issue?

matusvalo:
Hi @auvipy. No issue. I just found it when digging in source code of py-amqp.

PR #221, opened 2018-11-25
Fix basic_consume() with no consumer_tag provided

matusvalo:
This PR fixes #162 . This PR modifies `Channel.basic_consume()` that 
* consumer_tag is used not directly from parameter value but instead value returned by broker. This fixes issue when `consumer_tag` was not specified because "random" value generated by broker is used in this case. 
* `ValueError` is raised when `consumer_tag` is not specified and `nowait` is set to True, In this case we cannot know the value of consumer_tag since we are not waiting for response.
* When `consumer_tag` is provided and `nowait` is set to True, the value of `consumer_tag` is used from parameter.

This PR also adds some new Integration tests mainly connected to `basic_consume()` method

PR #240, opened 2019-01-14
basic_consume() should return consumer tag instead of tuple

### LINKED ISSUE #239: basic_publish returns a tuple instead of a `consumer_tag`
Hi,

Since 2.4, I notice the call to `basic_consume` without `consumer_tag` nor `nowait` now returns a tuple with a `consumer_tag` instead of the `consumer_tag` itself as was the case in the previous release.
It is not clear to me whether it's intended or not.

matusvalo:
This PR fixes change to API in #221. 

Closes #239


thedrow:
I'll release it today.

PR #245, opened 2019-01-16
When attempting to publish a message or start Celery worker amqp woul…

priordan83:
…d return "Unrecoverable error" KeyError.  Wrapped client connection properties in try catch given we pass on socket.timout.  This is not an issue on 2.3.2.

File "/opt/fs/prodtools-ia/dev/aspera-tools-client/venv/lib/python2.7/site-packages/amqp/channel.py", line 1755, in _basic_publish
    if client_properties['capabilities']['connection.blocked']:
KeyError: u'connection.blocked'


priordan83:
Already done in another pull request.

PR #247, opened 2019-01-24
Always treat SSLError timeouts as socket timeouts

dirkmueller:
Without specifically handling this case, the socket.timeout()
was not raised sometimes causing the connection to lock up.

In the case we hit the errno was None, so the previous if
condition did not apply.

### DIFF
--- a/amqp/transport.py
+++ b/amqp/transport.py
@@ -376,6 +376,10 @@ class SSLTransport(_AbstractTransport):
                 try:
                     s = recv(n - len(rbuf))  # see note above
                 except socket.error as exc:
+                    # ssl.sock.read may cause a SSLerror without errno
+                    # http://bugs.python.org/issue10272
+                    if isinstance(exc, SSLError) and 'timed out' in str(exc):
+                        raise socket.timeout()
                     # ssl.sock.read may cause ENOENT if the
                     # operation couldn't be performed (Issue celery#1414).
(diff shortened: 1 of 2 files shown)


dirkmueller:
TODO; needs test suite update.

aojea:
> TODO; needs test suite update.

https://github.com/dirkmueller/py-amqp/pull/1

matusvalo:
Thank you @dirkmueller. Could you update unittests. After updating them I will merge the PR but I prefer to do it after 2.4.1 release (to fix only issues introduced in 2.4.0).

matusvalo:
@dirkmueller I understand that this issue is **not** affecting python with version >= 3.2. Is it correct?

aojea:
> @dirkmueller I understand that this issue is **not** affecting python with version >= 3.2. Is it correct?

@matusvalo just curiosity, why shouldn't affect it?

dirkmueller:
@matusvalo I need to check this question, will take a bit.

matusvalo:
I was referring to the issue mentioned in the PR. It was not 100% clear from it whether the fix changed generic `SSLError` to `socket.timeout` or what was the fix. If it is true, I understand it in a way that this PR is not longer needed. If it is true it I would put a comment about it. It is not a big deal. I think we can merge this PR in any case and add comment later.

Issue #249, opened 2019-01-25
Increase code coverage in amqp/transport.py


aojea:
Current code coverage in amqp/transport.py is lower than the rest of the code and there are several functions that don't have any coverage that's very risky

```
Name                       Stmts   Miss Branch BrPart  Cover
------------------------------------------------------------
amqp/__init__.py              20      0      0      0   100%
amqp/abstract_channel.py      89      0     30      0   100%
amqp/basic_message.py         16      0      0      0   100%
amqp/channel.py              214      2     50      0    99%
amqp/connection.py           275      9     74      5    96%
amqp/exceptions.py            75      0      8      0   100%
amqp/method_framing.py        95      0     32      2    98%
amqp/platform.py              34      2     10      2    91%
amqp/protocol.py               4      0      0      0   100%
amqp/sasl.py                  90      4     16      1    93%
amqp/serialization.py        431      6    186      5    98%
amqp/spec.py                  81      0      0      0   100%
amqp/transport.py            280     34     94     12    84%
amqp/utils.py                 40      6      8      0    83%
------------------------------------------------------------
TOTAL                       1744     63    508     27    95%
```

PR #250, opened 2019-01-25
Add unit test for test_wrap_socket_sni

aojea:
Reference https://github.com/celery/py-amqp/issues/249


matusvalo:
Thank you @aojea

PR #251, opened 2019-01-25
Add unit test for SSLTransport _write function

aojea:
Reference:
https://github.com/celery/py-amqp/issues/249


t/unit/test_transport.py
```diff
@@ -600,6 +600,26 @@ def test_read_EOF(self):
                            match=r'.*Server unexpectedly closed connection.*'):
             self.t._read(64)
 
+    def test_write_success(self):
+        self.t.sock = Mock(name='SSLSocket')
+        self.t.sock.write.return_value = 2
+        self.t._write('foo')
```
matusvalo: Could you check here also that self.t.sock.write was called?

PR #254, opened 2019-02-07
Added support of cygwin platform

matusvalo:
Tested locally and now able co connect to remote RMQ broker.


matusvalo:
Tests are failing due coverage which cannot be increased of course :-)

PR #255, opened 2019-02-07
Fix spacing from #254.

clokep:
#254 added some spacing that didn't match the rest of the file, this just makes it consistent.


matusvalo:
Thank you @clokep. I missed wrong spacing.

clokep:
@matusvalo Not a problem! Glad to help out! 👍 I'm surprised that flake8 didn't find it to be honest.

PR #256, opened 2019-02-07
ssl client certs + SNI fixes

Gundersanne:


matusvalo:
@Gundersanne thank you for your PR. Is there any Issue which this PR is fixing? Is there any way how we can cover this PR with some unittests?

Gundersanne:
hey @matusvalo, I can look into writing a unit test somewhere this week. 

So ssl client certificates in combination with SNI (server_hostname is specified) is broken. On one part because it connects twice (and the first connct will fail, though you could avoid this with turning off automatic handshake I think), and on the other hand custom CA's aren't loaded with SNI.

I still had an issue with TCP connection being reset (regardless of heartbeat) over openshift/k8s clusters (connecting to the amqp queue on cluster A from cluster B). Pika didn't have this issue so not sure what's going on.

matusvalo:
Thank you @Gundersanne. Honestly, I don’t have much experience with ssl module. Currently, I am working on integration tests with Rabbit servers. Is it possible to test it directoly with real rabbit instance? But this is bigger effort since test needs to be set to configure tls enabled Rabbitmq broker. It should be duable because pika has something simmilar.

matusvalo:
Maybe @Gundersanne  it would be great if you could create issue with reference to this PR to have it tracked...

thedrow:
Please rebase this and add tests.

PR #261, opened 2019-03-01
Adds Python 3.7 to Classifiers

Coop56:
When I was researching on PyPi I noticed that Python 3.7 was missing from the list of supported language classifiers.  After digging in further I saw that Python 3.7 support was added in a previous commit.  This small PR just updates the supported languages for PyPi and the Appveyor configuration.


matusvalo:
Thank you @Coop56.

PR #264, opened 2019-03-09
Added hypothesis property based testing

thedrow:
This PR uses hypothesis to check if arrays are serialized and deserialized correctly.

### DIFF
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -1,3 +1,4 @@
 case>=1.3.1
 pytest>=3.0
 pytest-sugar>=0.9.1
+hypothesis>=4.8.0
--- a/t/unit/test_serialization.py
+++ b/t/unit/test_serialization.py
@@ -1,9 +1,10 @@
 from __future__ import absolute_import, unicode_literals
 
-from datetime import datetime
 from decimal import Decimal
+from datetime import datetime
 from math import ceil
 
+from hypothesis import given, reproduce_failure, strategies as st
 import pytest
 
 from amqp.basic_message import Message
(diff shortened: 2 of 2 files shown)


thedrow:
Yes, the examples directory should be ignored when reviewing this.

thedrow:
Seems like we have a genuine problem with strings inside dictionaries on Python 2.
Even without any adjustments that were made in this branch I can reproduce this:
```
Falsifying example: test_table(self=<t.unit.test_serialization.test_serialization instance at 0x7fdf17cde518>, table={u'': u'\x80'})
  File "/home/omer/Documents/Projects/py-amqp/t/unit/test_serialization.py", line 126, in test_table
    assert loads(b'F', dumps(b'F', [table]))[0][0] == table
AssertionError: assert {'': '\xc2\x80'} == {'': '\x80'}
  Differing items:
  {'': '\xc2\x80'} != {'': '\x80'}
  Full diff:
  - {'': '\xc2\x80'}
  ?       ----
  + {u'': u'\x80'}
  ?  +    +
```

thedrow:
Since `'\x80'` is a unicode string when you encode it to bytes, you get `\xc2\x80` but we aren't decoding back. We're encoding again.

Not sure how to resolve this issue.

thedrow:
Authentication fails with the addition of support for small strings in tables and arrays.
I'm not sure why though. 

Also, tox-docker doesn't seem to work anymore which is super strange.

thedrow:
Seems like there's a bug in RabbitMQ with short strings in tables.
It is allowed in the spec.

thedrow:
I found the source of the problem:
When RabbitMQ authenticates using the AMQPLAIN method it expects [a longstr instead of a shortstr](https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_auth_mechanism_amqplain.erl#L48).
See issue https://github.com/rabbitmq/rabbitmq-server/issues/1914.

PR #279, opened 2019-06-10
Basic publish confirm timeout

### LINKED ISSUE #276: Infinite wait when using confirm_publish
I am trying to get a robust amqp client and with this code I get an infinite wait in the last publish.

```python
from kombu import Connection, Producer, Queue
from amqp.exceptions import NotFound
import uuid
import os

c = Connection(amqp_url, transport_options={'confirm_publish': True})
p = Producer(c)
q = Queue('my-queue-{}'.format(uuid.uuid4()), channel=c)
q.declare()
p.publish({'foo': 'bar'}, routing_key=str(uuid.uuid4()), retry=False)

q.delete()
try:
    q.get()
    raise Exception("This should raise a NotFound")
except NotFound:
    q.declare()

p.publish({'foo': 'bar'}, routing_key=str(uuid.uuid4()), retry=False)

```

It seems there is no timeout property here https://github.com/celery/py-amqp/blob/09cf92a643ed075d048856a56a2e730482401558/amqp/channel.py#L1789

- kombu 4.5.0 
- pyamqp 2.4.2 

Any idea of what happens ? Thank you !

thedrow:
Fixes #276.

### DIFF
--- a/amqp/channel.py
+++ b/amqp/channel.py
@@ -1786,7 +1786,10 @@ class Channel(AbstractChannel):
             self.confirm_select()
         ret = self._basic_publish(*args, **kwargs)
         # Waiting for confirmation of message.
-        self.wait([spec.Basic.Ack, spec.Basic.Nack], callback=confirm_handler)
+        timeout = kwargs.get('timeout', None)
+        self.wait([spec.Basic.Ack, spec.Basic.Nack],
+                  callback=confirm_handler,
+                  timeout=timeout)
         return ret
 
     def basic_qos(self, prefetch_size, prefetch_count, a_global,
--- a/amqp/channel.py
+++ b/amqp/channel.py
@@ -1775,6 +1775,7 @@ class Channel(AbstractChannel):
     basic_publish = _basic_publish
 
     def basic_publish_confirm(self, *args, **kwargs):
+        confirm_timeout = kwargs.pop('confirm_timeout', None)
 
         def confirm_handler(method, *args):
             # When RMQ nacks message we are raising MessageNacked exception
@@ -1786,7 +1787,7 @@ class Channel(AbstractChannel):
             self.confirm_select()
         ret = self._basic_publish(*args, **kwargs)
         # Waiting for confirmation of message.
-        timeout = kwargs.get('timeout', None)
+        timeout = confirm_timeout or kwargs.get('timeout', None)
         self.wait([spec.Basic.Ack, spec.Basic.Nack],
                   callback=confirm_handler,
                   timeout=timeout)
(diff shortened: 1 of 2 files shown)


amqp/channel.py
```diff
@@ -1786,7 +1787,10 @@ def confirm_handler(method, *args):
             self.confirm_select()
         ret = self._basic_publish(*args, **kwargs)
         # Waiting for confirmation of message.
-        self.wait([spec.Basic.Ack, spec.Basic.Nack], callback=confirm_handler)
+        timeout = confirm_timeout or kwargs.get('timeout', None)
```
maingoh: `confirm_timeout` is not enough ? A more global timeout can be sent here ?

thedrow: You can either use the timeout for publishing or pass a specific timeout for waiting.
I'd rather avoid not having timeouts as much as possible.

maingoh: Okay, it LGTM ! I let @matusvalo approving

matusvalo:
I see only one problem. We are changing the semantics of `basic_publish_confirm()` because:
* **before**: with `timout=N` `wait()` in `basic_publish_confirm()` was not limited
* **after**: with `timeout=N` `wait()` in `basic_publish_confirm()` takes at most `N` seconds 

I am not against the change but I think we need to be careful. Moreover, I miss documentation of the new parameter. Otherwise LGTM.

thedrow:
So what should we do? Split the timeout in two?
Add a default (long) timeout for wait and allow to override it through `wait_timeout`?

matusvalo:
No my point was just be careful. I propose to add this change in release removing python 2 support (5.0 I suppose) and document it properly. In general I am fine with change.

RezaSi:
Any update on this? i have the same problem!
@matusvalo @thedrow

thedrow:
We need to address this.
@matusvalo Do you have time to complete this PR by any chance?

mehdy:
@thedrow @matusvalo This is been a top priority for us. Is there anything we could do to help with fixing this issue?
I'd really appreciate it if you could help us get this patch.

thedrow:
I, unfortunately, don't have the time right now.
You can take a look at the review comments, clone my branch, and complete this PR.
We'll get it released ASAP.

mehdy:
@thedrow Thank you! We'll do it!

mehdy:
@thedrow Just to be sure, as far as I understand this PR just needs some documentation of the new parameter with emphasis on the new behaviour it introduces. Right?

thedrow:
We also need to adjust the tests.

mehdy:
Sure! Thanks.

matusvalo:
Currently, I don't have neither much time. PRs are welkome!

PR #280, opened 2019-06-12
Ignore all methods except Close and Close-OK when channel/connection is closing

matusvalo:
This PR is adjusting closing Channel and Connection to conform AMQP 0.9.1 specification. Now after sending `Close()` method on channel (connection) all messages except channel (connection) `Close()` and `Close-OK()` are discarded.

### DIFF
--- a/amqp/connection.py
+++ b/amqp/connection.py
@@ -162,6 +162,10 @@ class Connection(AbstractChannel):
     }
     _METHODS = {m.method_sig: m for m in _METHODS}
 
+    _ALLOWED_METHODS_WHEN_CLOSING = (
+        spec.Connection.Close, spec.Connection.CloseOk
+    )
+
     connection_errors = (
         ConnectionError,
         socket.error,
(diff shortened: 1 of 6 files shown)


matusvalo:
@auvipy @thedrow  would you mind to review the PR?

matusvalo:
Apicheck is failing in CI. It seems not related to the PR.

amqp/connection.py
```diff
@@ -292,6 +297,21 @@ def _setup_listeners(self):
             spec.Connection.CloseOk: self._on_close_ok,
         })
 
+    def dispatch_method(self, method_sig, payload, content):
```
auvipy: ain't the dispatch method is same in both places?

matusvalo: They are not but the diff is not huge. They differ in methods that should be handled during closing:
```python
        if self.is_closing and method_sig not in (
            spec.Connection.Close, spec.Connection.CloseOk
        ):
```
vs. 
```python
        if self.is_closing and method_sig not in (
            spec.Channel.Close, spec.Channel.CloseOk
        ):
```

matusvalo: Maybe we can add list of method not ignored in channel and connection and move the logic to abstract channel. What do you think?

auvipy: sensible that would be

PR #281, opened 2019-06-21
If we don't want timeouts we must set the socket's timeout to None

thedrow:
See https://docs.python.org/3/library/socket.html#socket.socket.settimeout.

This commit was triggered by https://github.com/celery/celery/issues/4876#issuecomment-502922066.
It could fix the issue as well.


thedrow:
With this change, the integration tests hang on Travis.

@matusvalo Any idea why?

matusvalo:
@thedrow, see my comments in the patch. The problem is that when you are using publisher confirm, the publisher checks before publishing if broker sent a message to him. This relies on `timeout=0` which has moved semantics to block forever.

thedrow:
So what should we do to fix this issue exactly?

PR #283, opened 2019-07-03
Fix faulty ssl sni intiation parameters

dhananjaysathe:
We need to have 'do_handshake_on_connect' default to false due to the faulty code that executes before the sni options can be applied.

Also 
 context.check_hostname should not default to true this is a faulty default
 certfile,keyfile need to be non null before context.load_cert_chain(certfile, keyfile)

### DIFF
--- a/amqp/transport.py
+++ b/amqp/transport.py
@@ -317,7 +317,7 @@ class SSLTransport(_AbstractTransport):
 
     def _wrap_socket_sni(self, sock, keyfile=None, certfile=None,
                          server_side=False, cert_reqs=ssl.CERT_NONE,
-                         ca_certs=None, do_handshake_on_connect=True,
+                         ca_certs=None, do_handshake_on_connect=False,
                          suppress_ragged_eofs=True, server_hostname=None,
                          ciphers=None, ssl_version=None):
         """Socket wrap with SNI headers.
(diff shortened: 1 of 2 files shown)


amqp/transport.py
```diff
@@ -317,7 +317,7 @@ def _wrap_context(self, sock, sslopts, check_hostname=None, **ctx_options):
 
     def _wrap_socket_sni(self, sock, keyfile=None, certfile=None,
                          server_side=False, cert_reqs=ssl.CERT_NONE,
-                         ca_certs=None, do_handshake_on_connect=True,
+                         ca_certs=None, do_handshake_on_connect=False,
```
thedrow: Does this faulty behaviour occur only in Python 2.7?
If so, we can disable it on Python 2.7 and enable for the rest.

Also, can you please explain what do you mean? Why should we change the default?

amqp/transport.py
```diff
@@ -357,8 +357,10 @@ def _wrap_socket_sni(self, sock, keyfile=None, certfile=None,
                 hasattr(ssl, 'SSLContext')):
             context = ssl.SSLContext(opts['ssl_version'])
             context.verify_mode = cert_reqs
-            context.check_hostname = True
-            context.load_cert_chain(certfile, keyfile)
+            if cert_reqs != ssl.CERT_NONE:
```
thedrow: These changes look good to me.
It never hurts to be extra careful.

amqp/transport.py
dhananjaysathe: If this is passed here https://github.com/celery/py-amqp/blob/master/amqp/transport.py#L353 the default python ssl context socket tries to do a handshake before the following code that adds sni headers even executes. The same is true for https://github.com/celery/py-amqp/blob/master/amqp/transport.py#L362 that tries to connect if the flag is set. 

We anyway call this in our code here https://github.com/celery/py-amqp/blob/master/amqp/transport.py#L305 so its better not to pass this as true at this point.

dhananjaysathe:
bump

PR #287, opened 2019-08-12
Undeprecate auto_delete flag for exchanges.

### LINKED ISSUE #286: auto_delete flag for exchanges is deprecated since 2014
Hi,

Since 2014, setting the `auto_delete` flag for exchanges outputs the following warning:

> VDeprecationWarning: The auto_delete flag for exchanges has been deprecated and will be removed from py-amqp v1.5.0.

Current py-amqp version is v2.4.2 and `auto_delete` flag for exchanges is still there.

What is the rationale behind this deprecation? [This comment](https://github.com/celery/kombu/issues/296#issuecomment-32184244) from @ask seems to indicate that it was deprecated within amqp itself a while ago, but I'm unable to find any evidence of that (see [spec](https://www.rabbitmq.com/amqp-0-9-1-reference.html#exchange.declare.auto-delete)) and it was never dropped from RabbitMQ as far as I know.

If it was actually never deprecated or if it was undeprecated from amqp, is it possible to remove the warning? If it is indeed deprecated from amqp, is it possible to mention it in the deprecation message (with a source) and provide an alternative solution in docs or somewhere? Or just remove the flag (the warning has been there for long enough IMO).

Thank you in advance.

gilbsgilbs:
It was undeprecated from amqp.

See https://www.rabbitmq.com/amqp-0-9-1-errata.html#section_25

Fixes #286

### DIFF
--- a/amqp/channel.py
+++ b/amqp/channel.py
@@ -5,7 +5,6 @@ from __future__ import absolute_import, unicode_literals
 import logging
 import socket
 from collections import defaultdict
-from warnings import warn
 
 from vine import ensure_promise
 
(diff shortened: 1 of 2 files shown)


gilbsgilbs:
Thanks.

Issue #197, opened 2018-06-15
Built-in TypeError exception from Connection on_inbound_method()


juhamkilpinen:
There are several reports from the other projects getting this exception when using Connection class:

```
   File "/usr/lib/python2.7/site-packages/amqp/connection.py", line 472, in on_inbound_method
     return self.channels[channel_id].dispatch_method(
 TypeError: 'NoneType' object has no attribute '__getitem__'

```
See for example:
https://bugs.launchpad.net/oslo.messaging/+bug/1504725
https://github.com/celery/kombu/issues/857
https://github.com/celery/kombu/issues/463

Does this mean that Connection class is misused or should there be similar check in on_inbound_method() as there is in channel() for self.channels being None, for example:

```
def on_inbound_method(self, channel_id, method_sig, payload, content):
    if self.channels is None:
        raise RecoverableConnectionError('Connection already closed.')

```
I have seen this problem e.g. in OpenStack Pike when creating heat stacks and stack creation can fail for that TypeError exception. If I add above check it seems that client can re-connect and stack creation does not fail (when RecoverableConnectionError is thrown instead of TypeError).

thedrow:
I'm not sure how this happens.
I'll need to see a test case that doesn't involve kombu somehow.

juhamkilpinen:
Sorry I cannot provide a simplified test case to point the issue. I think it is somehow related to caching the connections with multiple process / threads involved, I.e. some kind of misuse of the API. I am happy for now that my workaround seems to work with the OpenStack env.

levsh:
Hi! I got the similar error.

      File "/opt/.../embedded/lib/python3.5/site-packages/celery/app/base.py", line 745, in send_task
        amqp.send_task_message(P, name, message, **options)
      File "/opt/.../embedded/lib/python3.5/site-packages/celery/app/amqp.py", line 552, in send_task_message
        **properties
      File "/opt/.../embedded/lib/python3.5/site-packages/kombu/messaging.py", line 181, in publish
        exchange_name, declare,
      File "/opt/.../embedded/lib/python3.5/site-packages/kombu/connection.py", line 494, in _ensured
        return fun(*args, **kwargs)
      File "/opt/.../embedded/lib/python3.5/site-packages/kombu/messaging.py", line 203, in _publish
        mandatory=mandatory, immediate=immediate,
      File "/opt/.../embedded/lib/python3.5/site-packages/amqp/channel.py", line 1743, in basic_publish_confirm
[...]

PR #289, opened 2019-09-11
Fix a channel issue against a connection already closed

4383:
In some possible case a connection can be closed (network issue, etc...) and
then channels reset to None, in this case if the client use
the `on_inbound_method` then we can facing a TypeError exception.

The `on_inbound_method` is not enough safer and then this method can
try to get channel from a none existing object.

In this case we need to raise a `RecoverableConnectionError` to allow client
to re-connect properly and re-init objects properly.

### DIFF
--- a/amqp/connection.py
+++ b/amqp/connection.py
@@ -510,9 +510,11 @@ class Connection(AbstractChannel):
         return self.on_inbound_frame(frame)
 
     def on_inbound_method(self, channel_id, method_sig, payload, content):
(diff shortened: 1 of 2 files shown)


4383:
Related to #197 I facing the same issue after some network issues

thedrow:
Can you please add a unit test to cover this?

thedrow:
Nevermind.
Got it.

PR #291, opened 2019-09-23
Separate logger for Connection.heartbeat_tick method

erinaceous:
I have a microservice=y project where I'm using py-amqp extensively. In the development environment I would guesstimate that 90% of the debug logs from these services are from `amqp.connection.Connection.heartbeat_tick`.

I still want to see all the other debug output from the `amqp` logger, but selectively filter out the heartbeat debug from my own code via
`logging.getLogger('amqp.connection.Connection.heartbeat_tick').setLevel(logging.INFO)`.


thedrow:
@auvipy Do we test logging at all?

PR #294, opened 2019-11-14
Fix pypi classifier development status

4383:
The version 5.0 is currently an alpha version
but pypi classifiers tag this version as a production/stable version,
this is wrong, this version is a pre-alpha version.

The main issue is that by using pip or requirements file to install
py-amqp it will pull and install the pre-alpha version (5.0) instead of
the stable version 2.5.2.

Normally pip don't install non stable version except if user
explicitly specify to install a non stable version.

Example with oslo.messaging:

```
$ pip install oslo.messaging
```

The previous command will pull py-amqp 5.0

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,7 @@ NAME = 'amqp'
 # -*- Classifiers -*-
 
 classes = """
-    Development Status :: 5 - Production/Stable
+    Development Status :: 2 - Pre-Alpha
     Programming Language :: Python
     Programming Language :: Python :: 2
     Programming Language :: Python :: 2.7


4383:
Related to https://review.opendev.org/#/c/693704/1

4383:
CI Errors don't seems related to my changes

4383:
Normally this version would be installed only if the `pip` `--pre` parameter would be given:

```
$ python -m pip install --pre amqp
```
Here this version is always pulled.

4383:
Also I guess that potentially pip need to take care of the classifiers and the release version based on semver to avoid this kind of situation.

auvipy:
you should also file an issue on pip I think

4383:
> you should also file an issue on pip I think

Yep I will.
Thanks

PR #299, opened 2020-01-07
Add support for TLS integration tests.

### LINKED ISSUE #270: Add integration tests for our SSL transport
Currently our integration tests do not verify our SSL support.
The container supports configuring SSL on the broker's side using environment variables.
However, the tox-docker doesn't support mounting or copying files which means we don't have a way to mount the required certificates.
We should use [trustme](https://github.com/python-trio/trustme) to generate the certificates before running the tests and copy the files into the image.

matusvalo:
This PR adds support for TLS integration tests against RabbitMQ instance. PR Adds custom RabbitMQ docker image with enabled TLS RabbitMQ. The TLS certificates used are self signed and are valid for 10 years (this should be enough :-) ).

Fixes #270


matusvalo:
@thedrow, @auvipy  there is one problem with this PR honestly. It seems that cpython 3.5 is unstable for integration stage :-(. It fails randomly with SIGSEGV, SIGABRT or SIGIOT. I am not sure how to fix it, but it seems that it is caused by python itself. All other python versions are fine and works all the time. I tried to use newer version of python (3.5.9) but nothing changed.

thedrow:
I don't see the problem in Travis though.
In any case, should we run the integration tests only with TLS?
I think we should be testing both.

matusvalo:
Hm honestly don't know where the problem can be, but I don't think is our code because as I said it seems that the issue is on C level:

```
Fatal Python error: GC object already tracked

Current thread 0x00007fd063114700 (most recent call first):

ERROR: InvocationError for command /home/travis/build/matusvalo/py-amqp/.tox/3.5-integration-rabbitmq/bin/py.test -xv -E rabbitmq t/integration -v (exited with code -6 (SIGIOT)) (exited with code -6)

3.5-integration-rabbitmq docker: remove 'f5b0692107' (forced)
```

 See the link below with failed test on my repo of py-amqp:

https://travis-ci.org/matusvalo/py-amqp/builds/633726919?utm_medium=notification&utm_source=github_status

Let me check it directly on my computer and I will see if it happens there. But anyways, we need to tackle it somehow if we don't want to have random CI failures. For now I see only to remove the 3.5 version from integration tests... Any other ideas?

Regarding the tests enabled I agree with you @thedrow we should have enabled both types of integration tests. In any case the patch is already executing both cases:

```python
    if request.param == 'plain':
        host = '%s:%s' % (
            os.environ.get('RABBITMQ_HOST', 'localhost'),
            os.environ.get('RABBITMQ_5672_TCP', '5672')
        )
        vhost = getattr(request.config, "slaveinput", {}).get("slaveid", None)
        return amqp.Connection(host=host, vhost=vhost)
    elif request.param == 'tls':
        host = '%s:%s' % (
            os.environ.get('RABBITMQ_HOST', 'localhost'),
            os.environ.get('RABBITMQ_5671_TCP', '5671')
        )
        vhost = getattr(request.config, "slaveinput", {}).get("slaveid", None)
        return amqp.Connection(
            host=host, vhost=vhost,
            ssl={
                'keyfile': 't/certs/client_key.pem',
                'certfile': 't/certs/client_certificate.pem'
            }
        )
```

matusvalo:
I have tried 3.5 python locally on my computer. The integration tests ran flawlessly there. Hence, I propose to try the PR and if there will be failures on 3.5 python I would exclude this python version from integration tests. Are you fine with this approach?

thedrow:
Yes, I agree.

PR #309, opened 2020-03-26
Fix amqp.Message docstring

### LINKED ISSUE #307: amqp.Message(timestamp:datetime)?
The documentation for the `timestamp` parameter of `amqp.Message`says it has to be of the type ` datetime.datetime`, see https://github.com/celery/py-amqp/blob/78ce51d5b4274fb2005b26f9d3e0bb4775b4d041/amqp/basic_message.py#L58-L59

When actually using a `datetime` here I get `error: cannot convert argument to integer` in https://github.com/celery/py-amqp/blob/78ce51d5b4274fb2005b26f9d3e0bb4775b4d041/amqp/serialization.py#L297

It seems that `dumps` expects `timestamp` to be go type `long long`.
(`amqplib` was converting `timestamp` via `time.mktime(val.timetuple())` to an int value.

Is my usage resp. understanding wrong or the documentation?

matusvalo:
Fixes #307

### DIFF
--- a/amqp/basic_message.py
+++ b/amqp/basic_message.py
@@ -55,7 +55,7 @@ class Message(GenericContent):
         message_id: shortstr
             The application message identifier
 
-        timestamp: datetime.datetime
+        timestamp: unsigned long
             The message timestamp
 
         type: shortstr


matusvalo:
unittests are failing due issue in pytest-sugar - PR fixing it: #310

PR #310, opened 2020-03-26
pin pytest==5.3.5 due issue in pytest-sugar

matusvalo:
Fixes issue in pytest-sugar: https://github.com/Teemu/pytest-sugar/issues/187

### DIFF
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -1,4 +1,4 @@
 case>=1.3.1
-pytest>=3.0
+pytest>=3.0,<=5.3.5
 pytest-sugar>=0.9.1
 pytest-rerunfailures>=6.0


auvipy:
well pinning to 5.x will break python 2. less then 5.4 should make them all happy

requirements/test.txt
```diff
@@ -1,4 +1,4 @@
 case>=1.3.1
-pytest>=3.0
+pytest>=3.0,<=5.3.5
```
auvipy: <=5.4?

matusvalo: With 5.4 pytest-sugar is not working - see: https://github.com/Teemu/pytest-sugar/issues/187

PR #258, opened 2019-02-26
Speed up serialization and deserialization using Rust

thedrow:
Current performance advantage is unclear as I'm aiming for feature parity at the moment.
I have added some benchmarks which we will use later on to measure how we're doing.

PR #311, opened 2020-03-30
Implement speedups in cython

matusvalo:
Cython speedups implementation

This PR is an alternative to #258.

PR consists of 3 parts

* Use __slots__ in GenericContent, since should be more lightweight than regular class
* Removes and simplifies parameters of functions in serialization.py - this is prerequisite for cythonizing + this should be faster in cpython 3.7+ [1] [2]
* simple cythonization of serialization.py hot paths using pure python mode - this enables two possible ways of using library a) cythonized b) not cythonized

Benefits:

* one single code base which is already well tested and tuned by users. This will introduce less maintenance burden comparing to #258
* one single language - lowers the barrier to contributors against #258
* low risk - by default is pure python mode. The speedups must be enabled by user when installing library. The speedups also shares the code with already proven code
* well tested framework - cython is regularly used for speedups: e.g. scipy [3], aiohttp [4], falcon [5]
* almost no change in source code - the PR is using .pxd "header" files so all cython specific code is outside the amqp code

Incompatible changes:
* cythonized code is not raising FrameSyntaxError due performance reasons
* distribution on Pypi must be changed from binary wheel to source only (sqlalchemy has the same approach)

What is pending:

- [x] performance tests (should be faster but no tests are done)
- [x] CI must be set to execute tests cythoned and pure python version of library
- [ ] Even more aggressive cythonization of serialization.py is possible needs some adjustments of serialization.py
- [x] ~~Distribution must be changed from binary wheel to source only~~
- [x] Installation must be documented

Installation:

* For regular python mode installation will be same as it is:
```
pip install amqp
```

* When user wants to install py-amqp with speedups (this is similar as it is done in sqlalchemy):
```
CELERY_ENABLE_SPEEDUPS=true pip install amqp
```

Running unittests:
* Unittests for pure python mode:
```
python setup.py test
```

* unittests for py-amqp with speedups:
```
CELERY_ENABLE_SPEEDUPS=true python setup.py test
```

[1] https://bugs.python.org/issue26110
[2] https://docs.python.org/3.7/whatsnew/3.7.html#cpython-bytecode-changes
[3] https://github.com/scipy/scipy/tree/master/scipy
[4] https://github.com/aio-libs/aiohttp/tree/master/aiohttp
[5] https://falconframework.org/#sectionFeatures


matusvalo:
@thedrow, @auvipy  please review this PR. I know that there is opened in parallel PR using Rust language and there are benefits of using Rust. But:
1. the progress is stalled and this PR is small and ready to go (almost - see What is pending).
2. I believe that implementing speedups has several benefits (at least in py-amqp project where number of contributors is limited) against Rust see Benefits section.

auvipy:
rustification will be postponed for future. cython shoulld be the go to go for now.

matusvalo:
@auvipy I will provide speed tests in next few days.

matusvalo:
@auvipy I ran simple test:
* pure python mode without speedups:
```
python -m timeit -s 'import amqp' "amqp.Message(body='Hello world', content_type='text/plain', application_content={'bar': 7, 'foo':5, 'list': [1,2,3]}, priority=5, correlation_id='correlation_id', reply_to='reply-to', expiration='expiration_value', message_id='message_id_value', timestamp=1234567890, type='type_value', user_id='user_id_value', app_id='app_id_value', cluster_id='cluster_id_value')._serialize_properties()"
10000 loops, best of 5: 35.5 usec per loop
```
* with speedups:
```
python -m timeit -s 'import amqp' "amqp.Message(body='Hello world', content_type='text/plain', application_content={'bar': 7, 'foo':5, 'list': [1,2,3]}, priority=5, correlation_id='correlation_id', reply_to='reply-to', expiration='expiration_value', message_id='message_id_value', timestamp=1234567890, type='type_value', user_id='user_id_value', app_id='app_id_value', cluster_id='cluster_id_value')._serialize_properties()"
10000 loops, best of 5: 20.1 usec per loop
```
So 20.1 usec with speedups against 35.5 usec without.

Moreover there is space to another improvements cythonizing `GenericContent` class:
```cython
cdef class GenericContent:
    cdef public object frame_method
    cdef public object frame_args
    cdef public object body
    cdef public list _pending_chunks
    cdef public int body_received
    cdef public int body_size
    cdef public bint ready
    cdef public dict properties

    @cython.locals(shift=cython.int, flag_bits=cython.int, flags=list)
    cpdef _serialize_properties(self)

    cpdef _load_properties(self, class_id, buf, offset)
```
In this case the test mentioned above is going to **15-17 usec** which is half value of pure python time.
But this is causing failing unittests and needs to be handled more carefully. I think it is candidate for next PR.

auvipy:
cool work and reasoning dude! you summed it up nicely which I failed due to my lack of good communication skills!

matusvalo:
@auvipy please do the final review. I have finished the PR. For documentation and installation I have created a followup PR: #313.

auvipy:
allow me sometime

auvipy:
do you think anything remains for this one?

matusvalo:
@auvipy for now not. I will follow up this PR with next improvements.

Issue #320, opened 2020-05-10
Framing errors on windows when using timeout due to BlockingIOError not handled right


timburke:
### High Level Summary

I've been tracking down an issue where using both the latest released stable and current `master` versions of `py-amqp` sometimes resulted in framing errors or py-amqp losing synchronization with the stream of frames coming from rabbitmq, causing it to hang because it misread data inside a message as the length of the next message.  

The bottom line turned out to be that there's an unhandled edge case inside `_AbstractTransport.read_frame` when used with a timeout on the read.  It checks for certain timeout errors in order to requeue partially read data but misses some exceptions that can be raised (at least on Windows).

I found that **sometimes** on Windows 10 at least, the `recv` call on a socket that is set with a timeout can raise `BlockingIOError`.  This does not cause data requeuing inside of `read_frame` which means that the next time someone calls `drain_events()` there will be 1 frame header (typically) chomped and the next frame will not be synchronized.

### Details

From the `read_frame` code:

```python
    def read_frame(self, unpack=unpack):
        read = self._read
        read_frame_buffer = EMPTY_BUFFER
        try:
            frame_header = read(7, True)
            read_frame_buffer += frame_header
            frame_type, channel, size = unpack('>BHI', frame_header)
            # >I is an unsigned int, but the argument to sock.recv is signed,
            # so we know the size can be at most 2 * SIGNED_INT_MAX
            if size > SIGNED_INT_MAX:
                part1 = read(SIGNED_INT_MAX)
                part2 = read(size - SIGNED_INT_MAX)
                payload = b''.join([part1, part2])
            else:
                payload = read(size)
            read_frame_buffer += payload
            ch = ord(read(1))
        except socket.timeout:
            self._read_buffer = read_frame_buffer + self._read_buffer
            raise
        except (OSError, IOError, SSLError, socket.error) as exc:  # <---- HERE IS THE PROBLEM
            # Don't disconnect for ssl read time outs
            # http://bugs.python.org/issue10272

            if isinstance(exc, SSLError) and 'timed out' in str(exc):  # <--- SECOND LATENT PROBLEM
                raise socket.timeout()
            if get_errno(exc) not in _UNAVAIL:
                self.connected = False
            raise
        if ch == 206:  # '\xce'
            return frame_type, channel, payload
        else:
            raise UnexpectedFrame(
                'Received {0:#04x} while expecting 0xce'.format(ch))
```

[...]

PR #321, opened 2020-05-17
Fix saving partial read buffer on windows during socket timeout.

timburke:
Fixes Issue #320.

When a frame read times out on Windows, the error raised did not trigger the transport logic to save off the initial partial read so that a future call to `read_frame()` would reread the partially read data.  This caused framing errors in certain cases on Windows.

Also fixes a similar error that could occur on any platform during reads on `len(payload) > SIGNED_INT_MAX`, where reading the payload is not an atomic read but split into two reads.  If the first succeeded but the second failed, the contents of the first read would be thrown away causing framing errors on subsequent calls to `read_frame`.

### DIFF
--- a/amqp/transport.py
+++ b/amqp/transport.py
@@ -2,6 +2,7 @@
 # Copyright (C) 2009 Barry Pederson <bp@barryp.org>
 from __future__ import absolute_import, unicode_literals
 
+import os
 import errno
 import re
 import socket
@@ -256,7 +257,15 @@ class _AbstractTransport(object):
             # so we know the size can be at most 2 * SIGNED_INT_MAX
             if size > SIGNED_INT_MAX:
                 part1 = read(SIGNED_INT_MAX)
-                part2 = read(size - SIGNED_INT_MAX)
+
+                # In case this read times out, we need to make sure to not
+                # lose part1 when we retry the read
+                try:
+                    part2 = read(size - SIGNED_INT_MAX)
+                except:
+                    read_frame_buffer += part1
+                    raise
+
                 payload = b''.join([part1, part2])
             else:
                 payload = read(size)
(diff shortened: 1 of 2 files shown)


amqp/transport.py
```diff
@@ -256,7 +257,15 @@ def read_frame(self, unpack=unpack):
             # so we know the size can be at most 2 * SIGNED_INT_MAX
             if size > SIGNED_INT_MAX:
                 part1 = read(SIGNED_INT_MAX)
-                part2 = read(size - SIGNED_INT_MAX)
+
+                # In case this read times out, we need to make sure to not
```
thedrow, line 261: Please move the comment to the except block as it belongs there.

amqp/transport.py
```diff
@@ -256,7 +257,15 @@ def read_frame(self, unpack=unpack):
...
+                # lose part1 when we retry the read
+                try:
+                    part2 = read(size - SIGNED_INT_MAX)
+                except:  # noqa
```
thedrow, line 265: Is this except too broad? Let's narrow it down if we can.

amqp/transport.py
```diff
@@ -266,10 +275,24 @@ def read_frame(self, unpack=unpack):
...
+            # errors.
+            if (
+                isinstance(exc, socket.error) and os.name == 'nt'
+                and get_errno(exc) == errno.EWOULDBLOCK  # noqa
+            ):
+                self._read_buffer = read_frame_buffer + self._read_buffer
+                raise socket.timeout()
+
             if isinstance(exc, SSLError) and 'timed out' in str(exc):
+                self._read_buffer = read_frame_buffer + self._read_buffer
```
thedrow, line 293: Can you leave a comment here too?

auvipy:
@matusvalo

matusvalo:
Give me some time for review.

auvipy:
sure please

matusvalo:
LGTM

PR #327, opened 2020-07-02
Stop to use deprecated method ssl.wrap_socket

### LINKED ISSUE #303: ssl.wrap_socket() is deprecated
`ssl.wrap_socket()` from https://github.com/celery/py-amqp/blob/78ce51d5b4274fb2005b26f9d3e0bb4775b4d041/amqp/transport.py#L353 is deprecated from python 3.7 (see [1]). Based on documentation `SSLContext.wrap_socket()` should be used instead.

[1] https://docs.python.org/3/library/ssl.html#ssl.wrap_socket

4383:
`ssl.wrap_socket` is deprecated since python 3.7 and since python 3.2
and 2.7.9 (released in 2014) it is recommended to use the
SSLContext.wrap_socket() instead of wrap_socket(). The top-level
function is limited and creates an insecure client socket without server
name indication or hostname matching [1].

Python 2.7 is now officially unmaintained, latest version of
python 2.7 is 2.7.18, py-amqp only support python versions who are compatible
with these changes [2].

These changes move away from `ssl.wrap_socket` by using
now `ssl.SSLContext.wrap_socket` [3].

Will close #303 

[1] https://docs.python.org/3/library/ssl.html#ssl.wrap_socket
[2] https://github.com/celery/py-amqp/blob/master/setup.py#L24,L29
[3] https://docs.python.org/3/library/ssl.html#ssl.SSLContext.wrap_socket

### DIFF
--- a/amqp/connection.py
+++ b/amqp/connection.py
@@ -94,8 +94,8 @@ class Connection(AbstractChannel):
     client name. For EXTERNAL authentication both userid and password are
     ignored.
 
-    The 'ssl' parameter may be simply True/False, or for Python >= 2.6
-    a dictionary of options to pass to ssl.wrap_socket() such as
+    The 'ssl' parameter may be simply True/False, or for Python >= 3.6
+    a dictionary of options to pass to ssl.SSLContext such as
     requiring certain certificates.
 
     The "socket_settings" parameter is a dictionary defining tcp
--- a/amqp/transport.py
+++ b/amqp/transport.py
@@ -336,39 +336,33 @@ class SSLTransport(_AbstractTransport):
 
     def _wrap_socket_sni(self, sock, keyfile=None, certfile=None,
                          server_side=False, cert_reqs=ssl.CERT_NONE,
-                         ca_certs=None, do_handshake_on_connect=False,
+                         do_handshake_on_connect=False,
                          suppress_ragged_eofs=True, server_hostname=None,
-                         ciphers=None, ssl_version=ssl.PROTOCOL_TLS):
+                         ssl_version=ssl.PROTOCOL_TLS):
         """Socket wrap with SNI headers.
 
-        Default `ssl.wrap_socket` method augmented with support for
+        stdlib `ssl.SSLContext.wrap_socket` method augmented with support for
         setting the server_hostname field required for SNI hostname header
         """
         opts = {
             'sock': sock,
-            'keyfile': keyfile,
-            'certfile': certfile,
             'server_side': server_side,
-            'cert_reqs': cert_reqs,
-            'ca_certs': ca_certs,
             'do_handshake_on_connect': do_handshake_on_connect,
             'suppress_ragged_eofs': suppress_ragged_eofs,
-            'ciphers': ciphers,
-            'ssl_version': ssl_version
+            'server_hostname': server_hostname,
         }
 
-        sock = ssl.wrap_socket(**opts)
+        context = ssl.SSLContext(ssl_version)
+        if certfile is not None:
+            context.load_cert_chain(certfile, keyfile)
+        if cert_reqs != ssl.CERT_NONE:
+            context.check_hostname = True
         # Set SNI headers if supported
         if (server_hostname is not None) and (
                 hasattr(ssl, 'HAS_SNI') and ssl.HAS_SNI) and (
                 hasattr(ssl, 'SSLContext')):
-            context = ssl.SSLContext(opts['ssl_version'])
             context.verify_mode = cert_reqs
-            if cert_reqs != ssl.CERT_NONE:
-                context.check_hostname = True
-            if (certfile is not None) and (keyfile is not None):
-                context.load_cert_chain(certfile, keyfile)
-            sock = context.wrap_socket(sock, server_hostname=server_hostname)
+        sock = context.wrap_socket(**opts)
         return sock
 
     def _shutdown_transport(self):
--- a/t/unit/test_transport.py
+++ b/t/unit/test_transport.py
@@ -1,7 +1,6 @@
 import errno
 import os
 import socket
-import ssl
 import struct
 from struct import pack
 from unittest.mock import ANY, MagicMock, Mock, call, patch
@@ -616,18 +615,13 @@ class test_SSLTransport:
 
     def test_wrap_socket_sni(self):
         sock = Mock()
-        with patch('ssl.wrap_socket') as mock_ssl_wrap:
+        with patch('ssl.SSLContext.wrap_socket') as mock_ssl_wrap:
             self.t._wrap_socket_sni(sock)
-            mock_ssl_wrap.assert_called_with(cert_reqs=0,
-                                             certfile=None,
-                                             keyfile=None,
-                                             sock=sock,
-                                             ca_certs=None,
+            mock_ssl_wrap.assert_called_with(sock=sock,
                                              server_side=False,
-                                             ciphers=None,
-                                             ssl_version=ssl.PROTOCOL_TLS,
+                                             do_handshake_on_connect=False,
                                              suppress_ragged_eofs=True,
-                                             do_handshake_on_connect=False)
+                                             server_hostname=None)
 
     def test_shutdown_transport(self):
         self.t.sock = None


4383:
I wondering if we should integrate these changes because py-amqp is in use in many projects that possibly use eventlet too and [eventlet have issues with recent versions of python and especially with the ssl module](https://github.com/eventlet/eventlet/issues/526).

[Fixes have been released in eventlet](https://github.com/eventlet/eventlet/pull/531) but the issue still occur.

Take a look to the example bellow. 

Copy the following code into a file (by example `/tmp/poc.py`):
```python
import argparse
import socket
import socketserver
import ssl

import eventlet


class MyServer(socketserver.BaseRequestHandler):

    def handle(self):
        # self.request is the TCP socket connected to the client
        self.data = self.request.recv(1024).strip()
        print("{} wrote:".format(self.client_address[0]))
        print(self.data)
        # just send back the same data, but upper-cased
        self.request.sendall(self.data.upper())


def client(host, port):
    context = ssl.create_default_context()
    with socket.create_connection((host, port)) as sock:
        with context.wrap_socket(sock, server_hostname=host) as ssock:
            print(ssock.version())


def main():
    parser = argparse.ArgumentParser()
    parser.add_argument('type', choices=['server', 'client'])
    parser.add_argument('--host', default='127.0.0.1')
    parser.add_argument('--port', default=8443)
    parser.add_argument('--patched', action='store_true')

    args = parser.parse_args()
    if args.patched:
        print("Monkey patching the stdlib")
        eventlet.monkey_patch()

    if args.type == 'server':
        with socketserver.TCPServer(
                (args.host, args.port), MyServer) as server:
            # Activate the server; this will keep running until you
            # interrupt the program with Ctrl-C
            server.serve_forever()
    else:
        client(args.host, args.port)


if __name__ == "__main__":
    main()
```

Then now run the following commands:

```sh
$ python3.8 -m venv /tmp/poc_venv
$ source /tmp/poc_venv/bin/activate
$ pip install eventlet
$ python /tmp/poc.py server & # start a fake server in background
$ python /tmp/poc.py client --patched
Monkey patching the stdlib
  File "/tmp/poc.py", line 50, in <module>
    main()
  File "/tmp/poc.py", line 46, in main
    client(args.host, args.port)
  File "/tmp/poc.py", line 23, in client
    with context.wrap_socket(sock, server_hostname=host) as ssock:
  File "/tmp/poc_venv/lib64/python3.8/site-packages/eventlet/green/ssl.py", line 437, in wrap_socket
    return GreenSSLSocket(sock, *a, _context=self, **kw)
  File "/tmp/poc_venv/lib64/python3.8/site-packages/eventlet/green/ssl.py", line 67, in __new__
    ret = _original_wrap_socket(
TypeError: wrap_socket() got an unexpected keyword argument '_context'
```
Potentially my proposed changes could introduce the same behavior seen in the previous example.

I'm not sure if we should take care of eventlet here but we can't ignore this context that is a real life example meet in many projects (openstack by example).

Why this is an issue for py-amqp? Because eventlet can monkey patch all the stdlib and so the ssl module too and it will pop up through py-amqp in this scenario. 

If we choose to continue with these changes we should be aware that projects that use eventlet in the parallel that they use py-amqp could facing this issue.

In same time this scenario stuck us to some python runtimes versions that still support `ssl.wrap_socket` and when stdlib will drop it's support then we should drop it here too and projects which use `py-amqp` and `eventlet` then will be impacted and then they will ought to choose to use stale python versions or to find solutions.

I think as soon we start to initiate this move as soon we will highlight this point and it will allow the community to centralize efforts to fix the eventlet side. Once these changes will be introduced if projects facing errors then they could cap their requirements to ignore issued versions.

thedrow:
I don't think we can drop support for eventlet right now.
You should attempt to help fix the issue there before we merge this pull request or include a workaround for eventlet in this pull request.
Otherwise, we won't be able to fix this issue.

amqp/transport.py
```diff
@@ -360,29 +360,25 @@ def _wrap_socket_sni(self, sock, keyfile=None, certfile=None,
...
         }
 
-        sock = ssl.wrap_socket(**opts)
+        context = ssl.SSLContext(ssl_version)
+        if certfile is not None:
+            context.load_cert_chain(certfile, keyfile)
+        if cert_reqs != ssl.CERT_NONE:
+            context.check_hostname = True
+        if (certfile is not None) and (keyfile is not None):
+            context.load_cert_chain(certfile, keyfile)
```
moisesguimaraes, line 375: This lines are redundant with the previous call to load_cert_chain a few lines above.

4383: Ah right good catch

4383:
> I don't think we can drop support for eventlet right now.
> You should attempt to help fix the issue there before we merge this pull request or include a workaround for eventlet in this pull request.

Indeed it could be an approach. My main concern is how to fix eventlet, personally I don't have much knowledge about the internal things of `eventlet` but I'm not against to dive into its code. My main goal here is mostly to reference this potential issue and ensure that we are aware of what could be happen with these changes.

> Otherwise, we won't be able to fix this issue.

4383:
> Please either find a workaround for eventlet or fix the original issue.
> We cannot merge this PR as it is.

Don't worry as said in my previous comment I not expect to see these changes merge, I mostly want to inform about this issue, I only fixed the @moisesguimaraes comment to keep the code clean and up-to-date.

You can consider these changes as some sort of WIP for now.

thedrow:
Can you try with master?
I was told https://github.com/eventlet/eventlet/commit/02fa826d22d0df7f9379c3a1e5cd2192e0ffadd8 should fix the issue.

thedrow:
> You can consider these changes as some sort of WIP for now.

I converted this PR to a draft.
Please let us know when it is ready for merge by turning it back to a PR again.

4383:
> Can you try with master?
> I was told [eventlet/eventlet@02fa826](https://github.com/eventlet/eventlet/commit/02fa826d22d0df7f9379c3a1e5cd2192e0ffadd8) should fix the issue.

Oh nice catch I don't seen this patch, I'll try to play a bit with it and see if it fix things.

4383:
> > You can consider these changes as some sort of WIP for now.
> 
> I converted this PR to a draft.
> Please let us know when it is ready for merge by turning it back to a PR again.

Sure make sense, thanks

thedrow:
Any news on this one?
Are you planning to complete it?
Eventlet 0.26.1 was released.

4383:
> Any news on this one?
> Are you planning to complete it?
> Eventlet 0.26.1 was released.

Not for the moment, I'll try to do some tests tomorrow.

4383:
Hello,

So after did some tests with eventlet 0.26.1, this version seens to fix the issue, I can't reproduce the issue described here with this version. This is a good news, so now I think we can continue with these changes.

You can use the script that I shared in my previous comment to do some tests too.

Also I think we can drop the "draft" tag of this PR.

Cheers

amqp/transport.py
```diff
@@ -343,7 +343,7 @@ def _wrap_socket_sni(self, sock, keyfile=None, certfile=None,
                          ciphers=None, ssl_version=None):
```
thedrow, line 343: Do we still need the arguments we removed from the [opts ](https://github.com/celery/py-amqp/pull/327/files#diff-e3b09014ce12269dff5eece1f9f7c082L361-L371) dictionary?
Some of them are unused and I think they either should be used or removed from the signature.

This is a private method so we can change its signature at will.

amqp/connection.py
```diff
@@ -94,8 +94,8 @@ class Connection(AbstractChannel):
     client name. For EXTERNAL authentication both userid and password are
     ignored.
 
-    The 'ssl' parameter may be simply True/False, or for Python >= 2.6
-    a dictionary of options to pass to ssl.wrap_socket() such as
+    The 'ssl' parameter may be simply True/False, or for Python >= 2.7
```
thedrow, line 97: Master supports 3.6+.

thedrow (changes requested):
Just a few minor touches.

thedrow:
@4383 Can you please rebase and resolve all the issues I pointed out in the code review?

4383:
> @4383 Can you please rebase and resolve all the issues I pointed out in the code review?

Ah yes sorry for the delay I was focusing on other stuffs.

amqp/transport.py
4383: Done, we can go ahead and drop them

amqp/connection.py
4383: Done

4383:
> @4383 Can you please rebase and resolve all the issues I pointed out in the code review?

Done

4383:
@thedrow the latest push correspond to a flake8 issue that I fixed, and the previous correspond to your comments addressed.

thedrow (approved):

thedrow:
Thank you so much!

4383:
You are welcome :)

PR #328, opened 2020-07-23
Fix buffer overflow in frame_writer after frame_max is increased

wedamija:
`frame_writer` allocates a `bytearray` on intialization with a length based on the
`connection.frame_max` value. If `connection.frame_max` is changed to a larger value, this causes an
error like `pack_into requires a buffer of at least 408736 bytes`.

From what I can tell, `connection._on_tune` can arbitrarily change this value, although I'm not
totally sure in what cases it actually occurs.

For context, we're upgrading from Celery 3.1.25 -> Celery 4.1, librabbitmq 1.6.1 -> amqp 2.6.0,
kombu 3.0.37 -> 4.2.2.post1. This has worked fine in a small production environment, but fails on
our main deployment. The main culprit we see are lots of errors raised from `frame_writer` with 
`pack_into requires a buffer of at least <x> bytes`, where x is always larger than our default 
`frame_max` of 131072. Despite this, `bigbody` is False, which ends up with us attempting to 
write a frame that is too large into the existing buffer. The only way I can see this happening is 
if `frame_max` increases via `connection._on_tune`.

### DIFF
--- a/amqp/method_framing.py
+++ b/amqp/method_framing.py
@@ -85,6 +85,20 @@ def frame_handler(connection, callback,
     return on_frame
 
 
+class Buffer(object):
+    def __init__(self, buf):
+        self.buf = buf
+
+    @property
+    def buf(self):
+        return self._buf
+
+    @buf.setter
+    def buf(self, buf):
+        self._buf = buf
+        self.view = memoryview(buf)
+
+
 def frame_writer(connection, transport,
                  pack=pack, pack_into=pack_into, range=range, len=len,
                  bytes=bytes, str_to_bytes=str_to_bytes, text_t=text_t):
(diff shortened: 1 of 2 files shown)


wedamija:
Thanks for approval @auvipy. How often do you release a new version of py-amqp?

auvipy:
expect one this month

amqp/method_framing.py
```diff
@@ -94,11 +108,17 @@ def frame_writer(connection, transport,
     # memoryview first supported in Python 2.7
     # Initial support was very shaky, so could be we have to
     # check for a bugfix release.
```
thedrow, line 110: This comment is probably out of place or even obsolete.
Do you mind issuing a new PR to remove it?

wedamija: Good catch, I'll put something up.

wedamija: @thedrow https://github.com/celery/py-amqp/pull/329. I think the best option is to just remove it since it seems likely to not be an issue at this point.

wedamija:
Just as a follow-up, we successfully deployed today using this fix.

auvipy:
will have a new release today.

PR #332, opened 2020-08-19
Use unittest.mock instead of case

matusvalo:


matusvalo:
Hmm seems that tests are failing for python 3.5 :-/

auvipy:
can you check the flake8

matusvalo:
flake8 fixed...

t/integration/test_integration.py
```diff
@@ -2,7 +2,10 @@
 from struct import pack
 
 import pytest
-from case import ANY, Mock, call, patch
+# python 3.5 does not support assert_called()
```
auvipy, line 5: master is 3.6+ only

auvipy (commented):
3.6+only should be ok

PR #347, opened 2020-11-26
Fix _wrap_socket_sni

moisesguimaraes:
* Change the default value of ssl_version to None. When not set, the
  proper value between ssl.PROTOCOL_TLS_CLIENT and ssl.PROTOCOL_TLS_SERVER
  will be selected based on the param server_side in order to create
  a TLS Context object with better defaults that fit the desired
  connection side.

* Change the default value of cert_reqs to None. The default value
  of ctx.verify_mode is ssl.CERT_NONE, but when ssl.PROTOCOL_TLS_CLIENT
  is used, ctx.verify_mode defaults to ssl.CERT_REQUIRED.

* Fix context.check_hostname logic. Checking the hostname depends on
  having support of the SNI TLS extension and being provided with a
  server_hostname value. Another important thing to mention is that
  enabling hostname checking automatically sets verify_mode from
  ssl.CERT_NONE to ssl.CERT_REQUIRED in the stdlib ssl and it cannot
  be set back to ssl.CERT_NONE as long as hostname checking is enabled.

* Refactor the SNI tests to test one thing at a time and removing some
  tests that were being repeated over and over.

Signed-off-by: Moisés Guimarães de Medeiros <guimaraes@pm.me>


JAORMX (approved):

auvipy (approved):

matusvalo (approved):
Thank you @moisesguimaraes. I looks good. I have just one question: Is the change breaking the current API?

moisesguimaraes:
@matusvalo I think the API is fine, but the travis tests are breaking on my fork. I was surprised that there was no error here due to that. My hypothesis is that the tests should load the certificate authority in order to validade the certificates coming from the docker container, the fail seems to be the evidence that client connections were not properly validating the server's identity.

matusvalo:
Thank you @moisesguimaraes . Can you provide also the fix of the integration tests?

moisesguimaraes:
Sure, I think I can spend some time on it tomorrow.

amqp/transport.py
```diff
@@ -506,20 +506,31 @@ def _wrap_socket_sni(self, sock, keyfile=None, certfile=None,
...
+                ssl.PROTOCOL_TLS_SERVER
+                if server_side
+                else ssl.PROTOCOL_TLS_CLIENT
+            )
+
         context = ssl.SSLContext(ssl_version)
+
         if certfile is not None:
             context.load_cert_chain(certfile, keyfile)
         if ca_certs is not None:
```
kgiusti, line 520: If ca_certs is None but verify_mode is != CERT_NONE it is worth calling SSLContext.set_default_verify_paths() to attempt to load the system-wide CA files.

kgiusti (commented):
I don't consider myself to be an SSL "expert" - more of an SSL "survivor".  Given that the patch LGTM.
Couple of comments in-line.

amqp/transport.py
```diff
@@ -506,20 +506,31 @@ def _wrap_socket_sni(self, sock, keyfile=None, certfile=None,
...
-        if ciphers:
+        if ciphers is not None:
             context.set_ciphers(ciphers)
-        if cert_reqs != ssl.CERT_NONE:
-            context.check_hostname = True
-        # Set SNI headers if supported
-        if (server_hostname is not None) and (
-                hasattr(ssl, 'HAS_SNI') and ssl.HAS_SNI) and (
-                hasattr(ssl, 'SSLContext')):
+        if cert_reqs is not None:
```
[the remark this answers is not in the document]
moisesguimaraes: this assumption is correct, the CI was failing basically because we were not providing the ca_certs to the client side.

4383 (approved):
LGTM

amqp/transport.py
moisesguimaraes: I will address this in a different patch.

4383:
> LGTM

Ah sorry I missed that Ken's comment isn't yet addressed

moisesguimaraes:
I need to address that in a different patch, as the right thing to do would be create a context via ssl.create_default_context() dropping the ssl_version param as we won't be using ssl.SSLContext() anymore.

thedrow:
Leaving this to you.

4383:
> I need to address that in a different patch, as the right thing to do would be create a context via ssl.create_default_context() dropping the ssl_version param as we won't be using ssl.SSLContext() anymore.

Ack I see

matusvalo (approved):

matusvalo:
Thank you @moisesguimaraes for your PR!

PR #348, opened 2020-12-13
[WIP] Abstract Types

auvipy:
pulling this of from 4+ years old PR so need to clean it up first, I am not expert in python typing yet so need more eyes before proceed further :D

### DIFF
--- a/amqp/types.py
+++ b/amqp/types.py
@@ -1,0 +1,5 @@
+from typing import Any, AnyStr, Mapping, Optional, Union
+
+SSLArg = Union[Mapping[AnyStr, Any], bool]
+MaybeDict = Optional[Mapping[AnyStr, Any]]
+Timeout = Optional[float]
--- a/amqp/types.py
+++ b/amqp/types.py
@@ -1,5 +1,566 @@
-from typing import Any, AnyStr, Mapping, Optional, Union
+"""Abstract types."""
+import abc
+import asyncio
+import socket
+from array import array
+from datetime import datetime
+from typing import (
+    Any, Awaitable, Callable, IO, List, Mapping,
+    MutableMapping, NamedTuple, Optional, Sequence, SupportsInt,
+    TypeVar, Tuple, Union,
+)
+from .protocol import queue_declare_ok_t
+from .spec import method_sig_t
+from vine import Thenable
 
-SSLArg = Union[Mapping[AnyStr, Any], bool]
-MaybeDict = Optional[Mapping[AnyStr, Any]]
-Timeout = Optional[float]
+Fd = TypeVar('Fd', int, IO)
+Int = TypeVar('Int', SupportsInt, str)
+
+
+class Frame(NamedTuple):
+    type: int
+    channel: int
+    data: bytes
+
+
+ConnectionBlockedCallbackT = Callable[[str], Optional[Awaitable]]
+ConnectionUnblockedCallbackT = Callable[[], Optional[Awaitable]]
+ConnectionFrameHandlerT = Callable[[Frame], Awaitable]
+ConnectionFrameWriterT = Callable[
+    [int,
+     int,
+     Optional[method_sig_t],
+     Optional[bytes],
+     Optional['MessageT'],
+     Optional[float]],
+    Awaitable,
+]
+
+WaitMethodT = Union[method_sig_t, Sequence[method_sig_t]]
+
+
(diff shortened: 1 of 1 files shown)


amqp/types.py
```diff
@@ -0,0 +1,566 @@
...
+    @abc.abstractmethod
+    async def __aexit__(self, *exc_info) -> None:
+        ...
+
+    @abc.abstractmethod
+    def _setup_listeners(self):
+        ...
+
+    @abc.abstractmethod
+    async def send_method(
```
auvipy, line 183: should we introduce async methods or start with sync only for now?

amqp/types.py
```diff
@@ -0,0 +1,566 @@
...
+    content_encoding: str
+    application_headers: MutableMapping
+    delivery_mode: int
+    priority: int
+    correlation_id: str
+    reply_to: str
+    expiration: str
+    message_id: str
+    timestamp: datetime
+    type: str
```
auvipy, line 130: this could be renamed to msg_type

amqp/types.py
```diff
@@ -0,0 +1,566 @@
+"""Abstract types."""
+import abc
+import asyncio
```
auvipy, line 3: should we just use asyncio for draft now or switch to anyio/trio?

amqp/types.py
```diff
@@ -0,0 +1,566 @@
...
+    Awaitable,
+]
+
+WaitMethodT = Union[method_sig_t, Sequence[method_sig_t]]
+
+
+class TransportT(metaclass=abc.ABCMeta):
+    """Transport type."""
+
+    rstream: asyncio.StreamReader
```
auvipy, line 45: I have to check the anyio/trio equivalent of these two

amqp/types.py
```diff
@@ -0,0 +1,566 @@
...
+    prev_recv: int
+
+    connection_errors: Tuple[type, ...]
+    channel_errors: Tuple[type, ...]
+    recoverable_connection_errors: Tuple[type, ...]
+    recoverable_channel_errors: Tuple[type, ...]
+
+    transport: TransportT
+    channels: MutableMapping[int, AbstractChannelT]
+    loop: asyncio.AbstractEventLoop
```
auvipy, line 270: anyio/trio alternatives?

amqp/types.py
thedrow: We should use anyio for generic clients.
This library is useful as a standalone as well.

amqp/types.py
thedrow: There is no alternative to the loop object in trio.
You should familiarize yourself with Task Groups (aka Nurseries in trio's terminology).

amqp/types.py
thedrow: This is a design question we need to answer before working on this PR.
As far as I'm concerned, we should drop support for synchronous connections as it will make it harder to maintain this library.

amqp/types.py
thedrow: I think it's clear as it is since the type is MessageT.

thedrow (changes requested):
I'm not sure why we're starting from this PR.
We should first make a plan of what needs to be done, document it in issues and start assigning them to contributors.

That way, we will be able to triage, track the progress of the implementation and distribute the work between us.

amqp/types.py
auvipy: OK. thanks

amqp/types.py
auvipy: btw, except https://github.com/agronholm/anyio/tree/master/docs docs I couldn't find another docsite.  thought you might have known?

auvipy:
> I'm not sure why we're starting from this PR.
> We should first make a plan of what needs to be done, document it in issues and start assigning them to contributors.
> 
> That way, we will be able to triage, track the progress of the implementation, and distribute the work between us.

My intention was to start playing with code, have some feedback, and make a CEP along with the coding experiment. now that I know about the thoughts of some of the core team members it will be easier for me to experiment and make a CEP draft and complete a basic POC and find out other things to be done[which could be done spliting]. I have a more clear direction now.

auvipy:
note: I didn't merge this in github! this commit somehow got merged in my local master branch which I didn't relase.

thedrow:
Can you please revert this?

auvipy:
> Can you please revert this?

I tried but GitHub dindt show any revert button so later removed them from master manually

PR #352, opened 2021-01-28
Fix CI appveyor/tox error

jorrit-wehelp:
tox-docker 2.0.0 was released and is backward incompatible.

Fix the tox ini section.


auvipy (approved):
thanks a lot!

PR #353, opened 2021-01-30
Pass long_description to setup()

flying-sheep:
Also gets rid of ugly `os.path` usage.

After my changes, we always use `here` as a reference point instead of using relative paths. Before, the code used

- `'./README.rst'` (in an uncalled function)
- `os.path.join(os.cwd(), p)` (which is the same as just using `p` directly)
- `os.path.join(here, p)` (which is relative to the directory `setup.py` is in)

Untested, please check before merging

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,9 +1,10 @@
 #!/usr/bin/env python
 
 import codecs
-import os
 import re
 import sys
+from os import environ
+from pathlib import Path
 
 import setuptools
 import setuptools.command.test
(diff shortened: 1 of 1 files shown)


auvipy:
I will check it locally first before merge

matusvalo (commented):
LGTM

flying-sheep:
soemone also needs to press the merge button :wink:

matusvalo:
Thank you @flying-sheep for your PR.

PR #357, opened 2021-02-18
Fix ordering of context.check_hostname and context.verify_mode

### LINKED ISSUE #356: [BUG] Cannot set cert_reqs=ssl.CERT_NONE due to order of context modification in _wrap_socker_sni
Setting the [broker_use_ssl](https://docs.celeryproject.org/en/stable/userguide/configuration.html#broker-use-ssl) value in celery configuration to `{"cert_reqs": ssl.CERT_NONE}` is not respected. This occurs because of the following snippet of code inside of `amqp.transport.SSLTransport._wrap_socket_sni`:

```
if cert_reqs is not None:
    context.verify_mode = cert_reqs
# Set SNI headers if supported
try:
    context.check_hostname = (
        ssl.HAS_SNI and server_hostname is not None
    )
except AttributeError:
    pass  # ask forgiveness not permission
```

Here we need to set `context.verify_mode` to the passed value of `cert_reqs` (in this case `ssl.CERT_NONE`). However, when `context.verify_mode = cert_reqs` executes it raises a `ValueError: Cannot set verify_mode to CERT_NONE when check_hostname is enabled.` This is because the default context object instantiated has `context.check_hostname=True` by default. The code above attempts to also set this value in the subsequent block:
```
try:
    context.check_hostname = (
        ssl.HAS_SNI and server_hostname is not None
    )
except AttributeError:
    pass  # ask forgiveness not permission
```
however, the exception has already been raised. These code blocks need to be revered in order for the verification to not occur (as per the `broker_use_ssl={"cert_reqs": ssl.CERT_NONE}` parameter. Setting `context.check_hostname=False` must occur before setting `context.verify_mode=ssl.CERT_NONE`. 

I'll submit a PR for review. Please let me know if I am missing some other security implication of this ordering!

Here's my basic app setup that creates the issue:

```
app = Celery(
    broker=amqps://...
)

app.conf.update( 
    broker_use_ssl={
        "cert_reqs": ssl.CERT_NONE,
    },
)
```

Which raises the noted `ValueError: Cannot set verify_mode to CERT_NONE when check_hostname is enabled.`

The use case is connecting to a RabbitMQ instance that runs behind a Traefik reverse proxy providing TLS termination. Since Traefik is dynamically provisioning certificates I do not want to perform client-side certificate verification.

The `ValueError` is swallowed by the celery worker somewhere, so the SSL verification ultimately fails (because the context is left with the default `context.verify_mode = ssl.CERT_REQUIRED` value), and the output the users see is the following:
```
[2021-02-18 21:24:33,507: ERROR/MainProcess] consumer: Cannot connect to amqps://admin:**@[**redacted**]:5671//: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate (_ssl.c:1076).
```

coltonbh:
Closes #356 

Please let me know if there are other security concerns to address here. This appears to be a bug since the code suggests that `context.verity_mode = ssl.CERT_NONE` and `context.check_hostname = False` should be a valid state.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -5,6 +5,17 @@ py-amqp is fork of amqplib used by Kombu containing additional features and impr
 The previous amqplib changelog is here:
 http://code.google.com/p/py-amqplib/source/browse/CHANGES
 
+.. _version-5.0.6:
+
+5.0.6
+=====
+- Change the order in which context.check_hostname and context.verify_mode get set
+  in SSLTransport._wrap_socket_sni. Fixes bug introduced in 5.0.3 where setting
+  context.verify_mode = ssl.CERT_NONE would raise
+  "ValueError: Cannot set verify_mode to CERT_NONE when check_hostname is enabled."
+  Setting context.check_hostname prior to setting context.verify_mode resolves the
+  issue.
+
 .. _version-5.0.5:
 
 5.0.5
--- a/amqp/transport.py
+++ b/amqp/transport.py
@@ -525,15 +525,24 @@ class SSLTransport(_AbstractTransport):
             context.load_verify_locations(ca_certs)
         if ciphers is not None:
             context.set_ciphers(ciphers)
-        if cert_reqs is not None:
-            context.verify_mode = cert_reqs
-        # Set SNI headers if supported
+        # Set SNI headers if supported. 
+        # Must set context.check_hostname before setting context.verify_mode 
+        # to avoid setting context.verify_mode=ssl.CERT_NONE while
+        # context.check_hostname is still True (the default value in context
+        # if client-side) which results in the following exception:
+        # ValueError: Cannot set verify_mode to CERT_NONE when check_hostname
+        # is enabled.
         try:
             context.check_hostname = (
                 ssl.HAS_SNI and server_hostname is not None
             )
         except AttributeError:
             pass  # ask forgiveness not permission
+
+        # See note above re: ordering for context.check_hostname and 
+        # context.verify_mode assignments.
+        if cert_reqs is not None:
+            context.verify_mode = cert_reqs
 
         if ca_certs is None and context.verify_mode != ssl.CERT_NONE:
             purpose = (
--- a/t/unit/test_transport.py
+++ b/t/unit/test_transport.py
@@ -700,7 +700,6 @@ class test_SSLTransport:
             set_ciphers_method_mock.assert_called_with(sentinel.CIPHERS)
 
     def test_wrap_socket_sni_cert_reqs(self):
-        # testing _wrap_socket_sni() with parameter cert_reqs == ssl.CERT_NONE
         with patch('ssl.SSLContext') as mock_ssl_context_class:
             sock = Mock()
             context = mock_ssl_context_class()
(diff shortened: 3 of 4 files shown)


coltonbh:
Also, this is a small contribution but if appropriate would be cool to add myself to the AUTHORS list. Been using `celery` for a long time--would be sweet to be an official contributor to part of it :)

auvipy:
@moisesguimaraes

auvipy (commented):
thanks for handling this. since changing the order fixes the issue, don't we need adjustment in tests as well? also a changelog entry is very much appreciated

coltonbh:
@auvipy Great! Happy to add a changelog entry. Re: tests--any pointers on exactly what you'd like done here? If you can point me to some tests that would exercise this specific piece of code I'd be happy to write a test or two. I'm not clear on the full context in which this class gets exercised.

auvipy:
related tests are located here https://github.com/celery/py-amqp/blob/master/t/unit/test_transport.py

coltonbh:
Great. I'll have a look and see what I can do. Thanks!

moisesguimaraes:
He y'all,

Thanks for the ping @auvipy.

I have something here to consider:

```
In [1]: import ssl

In [2]: context = ssl.SSLContext(ssl.PROTOCOL_TLS_SERVER)

In [3]: context.check_hostname
Out[3]: False

In [4]: context = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
   ...: 

In [5]: context.check_hostname
Out[5]: True
```

We should probably cover both scenarios with tests here.

moisesguimaraes:
A few more tests

```
In [29]: context = ssl.SSLContext(ssl.PROTOCOL_TLS_SERVER)

In [30]: context.verify_mode
Out[30]: <VerifyMode.CERT_NONE: 0>

In [31]: context.check_hostname = True

In [32]: context.verify_mode
Out[32]: <VerifyMode.CERT_REQUIRED: 2>

In [33]: context.verify_mode = ssl.CERT_NONE
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
```

I'm happy with the current changes, @coltonbh++.

matusvalo:
I am happy with changes. The only missing point are tests.

coltonbh:
@moisesguimaraes @auvipy @matusvalo Changelog entry added and tests added. I'm not quite sure what to assert on the tests since the previous tests mistakenly passed due to the entire `context` object being mocked. Here my goal is to leave as much of the `context` object's default behavior intact and assure than the code can execute successfully. Do you have further suggestions on how to improve these tests? Thanks!

matusvalo:
@coltonbh the one possible solution how test properly this is to test it via integration tests:

https://github.com/celery/py-amqp/blob/master/t/integration/test_rmq.py

auvipy (approved):

coltonbh:
@matusvalo I'm trying to test that the modifications on the `context` object performed in `_wrap_socket_sni` do not raise exceptions (as previously they did). My tests do fail under the old code (at least the `ssl.PROTOCOL_TLS_CLIENT` version does since it sets `check_hostname=True`). After my question above I added the `mock_wrap_socket.assert_called_once()` check which assures that the code block executed without error and called the `wrap_socket()` method. Do we want to add integration tests? I'm not sure what it would add to actually connect to a RabbitMQ instance since that wasn't the original problem. Please let me know if these feel urgent--I do believe the current tests--with my final assert statement--cover the bug :)

moisesguimaraes:
Let me have a quick look

moisesguimaraes:
I would move your tests to the end of the function and mock both `load_default_certs` and `wrap_socket` as these would be the expensive calls. So first we test the code path, then we test if `ssl.SSLContext` behaves accordingly.

I would favor tweaking `server_side` param instead of `ssl_version` to set the side of the connection.

Let's remember that checking the hostname with SNI is a client side feature only. There is no way for servers to validate a client's hostname as they don't have one. Focus the test on exercising the possibility to set cert_reqs to both `ssl.CERT_NONE` and `ssl.CERT_REQUIRED` in both scenarios (client side and server side), in other words the test should cover the 4 scenarios of their combinations.

Having the ssl.SSLContext constructor not mocked should fail with the old code and work with the new one. I don't see the need for integration tests at this point.

Looking again at your code, you're almost there. I strongly consider mocking `load_default_certs` as it is an expensive call when loading all system certs.

Thanks again for your contribution @coltonbh o/

Changelog
```diff
@@ -4,6 +4,16 @@ Changes
 py-amqp is fork of amqplib used by Kombu containing additional features and improvements.
 The previous amqplib changelog is here:
 http://code.google.com/p/py-amqplib/source/browse/CHANGES
+.. _version-5.0.6:
```
thedrow, line 7:
```suggestion

.. _version-5.0.6:
```

amqp/transport.py
```diff
@@ -535,6 +533,9 @@ def _wrap_socket_sni(self, sock, keyfile=None, certfile=None,
         except AttributeError:
             pass  # ask forgiveness not permission
 
+        if cert_reqs is not None:
```
thedrow, line 536: If you don't mind, let's document inline what each condition is meant to do and why their order is this way and not another.
Otherwise this issue may regress.

t/unit/test_transport.py
```diff
@@ -700,7 +700,22 @@ def test_wrap_socket_ciphers(self):
             set_ciphers_method_mock.assert_called_with(sentinel.CIPHERS)
 
     def test_wrap_socket_sni_cert_reqs(self):
-        # testing _wrap_socket_sni() with parameter cert_reqs == ssl.CERT_NONE
+        # testing context creation inside _wrap_socket_sni() with parameter 
+        # cert_reqs == ssl.CERT_NONE
+        with patch('ssl.SSLContext.wrap_socket') as mock_wrap_socket:
+            sock = Mock()
+            self.t._wrap_socket_sni(
```
thedrow, line 707: Usually, we don't test the internals of a class.
We exercise the public API and check the outcomes.
This is because this test may pass but another code path in the public method will still make the desired functionality unobtainable.

thedrow (changes requested):
I agree that integration tests will help us cover this case.
Overall, this is good work. See my comments.

Changelog
coltonbh: Fixed!

coltonbh:
ok @moisesguimaraes and @thedrow your suggestions have been incorporated. Specifically:

- Added the additional tests suggested by @moisesguimaraes so the the full 2x2 matrix of possibilities is covered
- Mocked out `load_default_certs` and asserted correct behavior for this call
- Moved tests to the end of the `test_wrap_socket_sni_cert_reqs` code block
- Added comments inline to clarify why the code must execute in a particular path.

Let me know if anything else is needed!

matusvalo (approved):

coltonbh:
🎉🥳

PR #343, opened 2020-11-07
Basic publish confirm timeout fix

RezaSi:

PR #360, opened 2021-05-06
Backport basic publish confirm fix to 2.6

ccorbacho:
This backports https://github.com/celery/py-amqp/pull/343 to the 2.6 branch, as Celery 4 is still supported until August.

We are seeing problem with Amazon RabbitMQ and publish confirm requests hanging, and the Gunicorn timeout is killing the process, rather than the retries kicking in (for non Gunicorn processes, we kill them if they block for too long, but this is not ideal).

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 2.6.1
+current_version = 2.6.2
 commit = True
 tag = True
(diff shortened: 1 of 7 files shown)


thedrow:
I'll deal with this once we release 5.0.

auvipy (approved):
merging for anyone using old version needs this for a smoother migration for instance.

PR #367, opened 2021-11-06
new pythons to actions

auvipy:


auvipy:
will need tox-gh-actions in tox for parsing 3.10

PR #368, opened 2021-11-18
adding experimental __slots__ to some classes

auvipy:
will add more slots and fix the issues which break the CI

related https://github.com/celery/vine/pull/67

### DIFF
--- a/amqp/basic_message.py
+++ b/amqp/basic_message.py
@@ -111,10 +111,6 @@ class Message(GenericContent):
         "delivery_info",
         "body",
         "channel",
-
-        # adding '__dict__' to get dynamic assignment
-        "__dict__",
-        "__weakref__",
         )
 
     @property
--- a/amqp/basic_message.py
+++ b/amqp/basic_message.py
@@ -107,6 +107,16 @@ class Message(GenericContent):
         self.body = body
         self.channel = channel
 
+    __slots__ = (
+        "delivery_info",
+        "body",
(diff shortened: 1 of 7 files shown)


amqp/basic_message.py
```diff
@@ -107,6 +107,16 @@ def __init__(self, body='', children=None, channel=None, **properties):
         self.body = body
         self.channel = channel
 
+    __slots__ = (
+        "delivery_info",
+        "body",
+        "channel",
+
+        # adding '__dict__' to get dynamic assignment
```
auvipy, line 115: this are redundant

auvipy:
good to see tests are passing

auvipy:
I want to include it on 5.1 release, may be after christmas or on the new years

auvipy:
image
load time without this patch/patches

auvipy:
@4383 as far i can recall you use this package on old openstack versions right? would love to have you try this changes

matusvalo (commented):
I don't see any benefit having `__slots__` for classes except `Message` class. `__slots__` is mainly performance related and the only class that is created on bigger scales is `Message` class.

auvipy:
IMHO, other classes when used in production, could be also benefited as well?

auvipy (commented):
I am merging this as I can see most of the classes are used by kombu and celery, so in bigger system tiny resource saving would be beneficial, IMHO

PR #374, opened 2021-12-12
Bugfix: not closing socket after server disconnect

michael-lazar:
When the connection is closed on the remote side, calling ``socket.shutdown()`` may raise an OSError. This fix makes sure that sock.close() is always called so the socket is closed cleanly.

https://stackoverflow.com/a/598759

See https://github.com/celery/celery/issues/4843#issuecomment-991394349 for more details.

PR #377, opened 2021-12-14
reduce memory usage of Connection

pawl:
While looking into https://github.com/celery/celery/issues/4843#issuecomment-991394349, I noticed [Connection._avail_channel_ids](https://github.com/celery/py-amqp/blob/v5.0.7/amqp/connection.py#L270) seems to be the thing that uses the most memory on a `Connection`:

Currently `_avail_channel_ids` is defined on `Connection` like this:
```
self._avail_channel_ids = array('H', range(self.channel_max, 0, -1))
```

It creates an array of all possible channel ids and removes them as they're used.

This PR reduces the memory usage of the `Connection` by reversing that logic and turning `_avail_channel_ids` into `_used_channel_ids`. This will only store the channel ids that are currently used, which is almost always going to be a much smaller array.

You can see the empty arrays use a lot less memory:
```
>>> from array import array
>>> import sys
>>> sys.getsizeof(array('H'))
64
>>> sys.getsizeof(array('H', range(65535, 0, -1)))
133524
```

There's downside though: It's going to take a little bit longer to `_get_free_channel_id`, because it's now going to start counting from the 1 to channel_max and return the first number that doesn't exist in `_used_channel_ids`.

However, this could potentially reduce the severity of memory leaks.

### DIFF
--- a/amqp/channel.py
+++ b/amqp/channel.py
@@ -150,7 +150,7 @@ class Channel(AbstractChannel):
         connection, self.connection = self.connection, None
         if connection:
             connection.channels.pop(channel_id, None)
-            connection._avail_channel_ids.append(channel_id)
+            connection._used_channel_ids.remove(channel_id)
         self.callbacks.clear()
         self.cancel_callbacks.clear()
         self.events.clear()
--- a/amqp/connection.py
+++ b/amqp/connection.py
@@ -267,7 +267,7 @@ class Connection(AbstractChannel):
         self.on_unblocked = on_unblocked
         self.on_open = ensure_promise(on_open)
 
-        self._avail_channel_ids = array('H', range(self.channel_max, 0, -1))
+        self._used_channel_ids = array('H')
 
         # Properties set in the Start method
         self.version_major = 0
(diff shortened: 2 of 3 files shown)


auvipy (commented):
thanks for your patch, you were involved in https://github.com/celery/py-amqp/pull/374/files aswell right?

auvipy:
Integrations tests are failing though so need some adjustment. Also I am almost done with https://github.com/celery/py-amqp/pull/368 can you guys test it on your setup

pawl:
> thanks for your patch, you were involved in https://github.com/celery/py-amqp/pull/374/files aswell right?

Yeah, I helped reproduce the issue.

> Integrations tests are failing though so need some adjustment.

I'm thinking this should fix it: https://github.com/celery/py-amqp/pull/377/commits/64f4900f91f31d8dfdb0a078023654e061469c31

> Also I am almost done with #368 can you guys test it on your setup

I can try running our unit tests with it. Our last release of the year is tomorrow, so we couldn't put that into production for a while. However, we will be sending the fix for #374 to production tomorrow.

auvipy:
OK thank you. you can try the slot related code in your local env as well.

auvipy (approved):

PR #385, opened 2021-12-21
improve performance of _get_free_channel_id, fix channel max bug

pawl:
This PR fixes a performance issue introduced by #377.

I noticed while running tests that `_get_free_channel_id` is way slower than I expected.

Surprisingly, this code that's similar to `_get_free_channel_id` (when all the channel ids are used) takes almost 40 seconds to run:
```
import time
from array import array

channel_max = 65535
used_channel_ids = array('H', range(1, channel_max + 1))
start = time.time()
for channel_id in range(1, channel_max + 1):
    if channel_id not in used_channel_ids:
        assert False
print(f'finished {time.time() - start:0.3f}')
```

This PR casts `_used_channel_ids` to a `set` before the loop, and it only takes ~0.004 seconds to run the loop after this change. It also allows keeping the memory usage improvements by keeping the `_used_channel_ids` stored as an array on the class.

I also learned that there's a bug preventing `_get_free_channel_id` from reaching the `channel_max` (the limit is currently channel_max - 1), and I fixed that too.

### DIFF
--- a/amqp/connection.py
+++ b/amqp/connection.py
@@ -482,8 +482,11 @@ class Connection(AbstractChannel):
         self._transport = self.connection = self.channels = None
 
     def _get_free_channel_id(self):
-        for channel_id in range(1, self.channel_max):
-            if channel_id not in self._used_channel_ids:
+        # Cast to a set for fast lookups, and keep stored as an array for lower memory usage.
+        used_channel_ids = set(self._used_channel_ids)
+
+        for channel_id in range(1, self.channel_max + 1):
+            if channel_id not in used_channel_ids:
                 self._used_channel_ids.append(channel_id)
                 return channel_id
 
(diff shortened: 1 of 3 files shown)


auvipy (changes requested):
can you also adjust some integration tests as well?

pawl:
@auvipy Is this what you were thinking?: https://github.com/celery/py-amqp/pull/385/commits/3db6df1b35f1cc4f213ed6d5f4e461c89d00c48e

auvipy:
yes

auvipy (approved):

skinnyh:
Hi, can we get a version bump for this fix? Thank you!

auvipy:
Yes of course, this weekend for sure

skinnyh:
> Yes of course, this weekend for sure

Hi, any update on it?

GuardianRain:
Hi, which version has already include this fix? Thank you!

auvipy:
5.1.1 please

PR #405, opened 2023-02-28
Send heartbeat frames more often

arnaudmorin:
The AMQP protocol is saying that we should send "two" "heartbeat frames" during the "heartbeat timeout" (see [1] and rabbit implement this in [2]).

The "two" value is the "rate" parameter in the current implementation.

The current implementation was sending only one frame during the "heartbeat timeout", which is wrong.

[1] https://www.amqp.org/specification/0-9-1/amqp-org-download
[2] https://www.rabbitmq.com/heartbeats.html#heartbeats-interval

### DIFF
--- a/amqp/connection.py
+++ b/amqp/connection.py
@@ -724,12 +724,17 @@ class Connection(AbstractChannel):
             once per second.
 
         Keyword Arguments:
-            rate (int): Previously used, but ignored now.
+            rate (int): Number of heartbeat frames to send during the heartbeat
+                        timeout
         """
         AMQP_HEARTBEAT_LOGGER.debug('heartbeat_tick : for connection %s',
                                     self._connection_id)
         if not self.heartbeat:
             return
+
+        # If rate is wrong, let's use 2 as default
+        if rate <= 0:
+            rate = 2
 
         # treat actual data exchange in either direction as a heartbeat
         sent_now = self.bytes_sent
@@ -755,7 +760,7 @@ class Connection(AbstractChannel):
         self.prev_sent, self.prev_recv = sent_now, recv_now
 
         # send a heartbeat if it's time to do so
-        if now > self.last_heartbeat_sent + self.heartbeat:
+        if now > self.last_heartbeat_sent + self.heartbeat / rate:
             AMQP_HEARTBEAT_LOGGER.debug(
                 'heartbeat_tick: sending heartbeat for connection %s',
                 self._connection_id)
--- a/tox.ini
+++ b/tox.ini
@@ -18,7 +18,7 @@ deps=
 sitepackages = False
 recreate = False
 commands =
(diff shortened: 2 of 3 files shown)


pslestang:
From my POV there is a main issue in the library implementation. The current implementation is mixing heartbeat (the interval between two thick) and the heartbeat timeout on amqp connection.

https://github.com/celery/py-amqp/blob/9d7b36a0da963b5097379e2aacb51e3fbd1ed06a/amqp/connection.py#L443

As written in rabbitmq documentation "It is important to not confuse the timeout value with the interval one".

The amqp connection should be open with the heartbeat timeout taking care of timeout negotiation with server and then send a tick every heartbeat interval.

Technically it is enough to get the heartbeat timeout and the heartbeat rate to compute the heartbeat interval.

amqp/connection.py
```diff
@@ -754,7 +754,7 @@ def heartbeat_tick(self, rate=2):
         self.prev_sent, self.prev_recv = sent_now, recv_now
 
         # send a heartbeat if it's time to do so
-        if now > self.last_heartbeat_sent + self.heartbeat:
+        if now > self.last_heartbeat_sent + self.heartbeat / rate:
```
auvipy, line 757: we need to verify this change with unit test

auvipy (changes requested):

amqp/connection.py
arnaudmorin: Also, I need at least to check the divide by 0 and remove the comment about deprecated parameter

arnaudmorin: done

auvipy:
I restarted

arnaudmorin:
up

tox.ini
```diff
@@ -18,7 +18,7 @@ deps=
 sitepackages = False
 recreate = False
 commands =
-    unit: py.test -xv --cov=amqp --cov-report=xml --no-cov-on-fail t/unit
+    unit: py.test -xv --cov=amqp --cov-report=xml --no-cov-on-fail t/unit {posargs}
```
auvipy, line 21:
```suggestion
    unit: pytest -xv --cov=amqp --cov-report=xml --no-cov-on-fail t/unit {posargs}
```

auvipy (changes requested):

arnaudmorin:
recheck

arnaudmorin:
I agree with what Pierre Samuel is explaining, but refactoring this would lead to an API change (the parameter would be renamed / reconfigured), and I dont think this is something we want on py-amqp now.

Eventually this could be done in another change.

auvipy:
can you please try to change the rabbitmq docker image to fix the integration test?

auvipy:
lets see how this https://github.com/celery/py-amqp/pull/414 go

auvipy:
you  can rebase now

PR #408, opened 2023-03-13
Update py versions in ci.yaml

auvipy:
pypy-3.9-unit: failed with env name pypy-3.9-unit conflicting with base python python3.9
  pypy-3.9-unit: FAIL code 1 (0.02 seconds)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


auvipy:
need to find a way to run pypy in the CI

Nusnus:
> need to find a way to run pypy in the CI

I suspect the issue is in the `tox.ini` file.
Compare it to `celery`'s, you'll see some differences regarding pypy.

auvipy:
celeryy do not use tox-docker like py-amqp and kombu. I will try to figure what is going wrong

PR #414, opened 2023-06-13
Update tox.ini rabbitmq docker image

auvipy:


arnaudmorin:
sounds not working as well

PR #432, opened 2024-05-28
Hard-code requests version because of the bug in docker-py

awmackowiak:
Due to [bug reported in decker-py](https://github.com/docker/docker-py/issues/3256) there is a need to hard-code the requests version to <=2.31.0.
5 days ago they released the newest version and the integration tests started failures that day.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -5,7 +5,9 @@ envlist =
     flake8
     apicheck
     pydocstyle
-requires = tox-docker>=2.0
+requires =
+    tox-docker<=4.1
+    requests<=2.31.0
 
 [testenv]
 deps=


auvipy (approved):

Nusnus:
The bug is fixed in docker-py 7.1.0.
I think it's better to upgrade that instead of limiting requests instead.

Nusnus:
The bug is fixed in docker-py 7.1.0.
I think it's better to upgrade that instead of limiting requests instead 🙏 @awmackowiak

Nusnus:
> The bug is fixed in docker-py 7.1.0.
> I think it's better to upgrade that instead of limiting requests instead 🙏 @awmackowiak 

This is because newer versions of requests will never fix it so we either limit requests forever or just use the newest docker-py

Nusnus:
> > The bug is fixed in docker-py 7.1.0.
> > I think it's better to upgrade that instead of limiting requests instead 🙏 @awmackowiak
> 
> This is because newer versions of requests will never fix it so we either limit requests forever or just use the newest docker-py

I am getting confused 4real…
I’ve tried to follow up on my own comment: https://github.com/celery/py-amqp/pull/433
Nothing works, not even close.

I don’t like limiting requests, but we might not have any other choice at the moment, especially if this issue is relevant to fixing the Redis bug.

Nusnus:
> > > The bug is fixed in docker-py 7.1.0.
> > > I think it's better to upgrade that instead of limiting requests instead 🙏 @awmackowiak
> > 
> > 
> > This is because newer versions of requests will never fix it so we either limit requests forever or just use the newest docker-py
> 
> I am getting confused 4real… I’ve tried to follow up on my own comment: #433 Nothing works, not even close.
> 
> I don’t like limiting requests, but we might not have any other choice at the moment, especially if this issue is relevant to fixing the Redis bug.

I also tried it with kombu, juts for the sake of seeing what happens.
Same confusing results: https://github.com/celery/kombu/pull/2010

Nusnus (approved):
> Existing custom
HTTPAdapters will need to migrate their code to use this new API.
get_connection is considered deprecated in all versions of Requests>=2.32.0.

This is probably broken in more places and will take some time to fix across the board.

CleanShot 2024-05-29 at 20 03 41

Requests 2.32+ appears to have had a bumpy road regardless:
CleanShot 2024-05-29 at 20 06 02

Taking into consideration everything, I give it my approval.
That being said, we will need to revisit this issue again to allow upgrading everything.

@awmackowiak @auvipy

awmackowiak:
@Nusnus 
I also tried to find a silver bullet for that problem but didn't find an easy way to get the test up and running.
Kudos for the try :)

Nusnus:
> Taking into consideration everything, I give it my approval.
> That being said, we will need to revisit this issue again to allow upgrading everything.

It looks like it still requires more attention #442

PR #449, opened 2025-07-06
Remove Python 3.8 support

kajinamit:
Python 3.8 reached its EOL on 2024-10-07[1].

[1] https://devguide.python.org/versions/


kajinamit:
OK. I don't argue that we should merge this now and I'll leave the timing for you.

However I wonder if we should update the classifiers and bump the minimum at least to 3.8 (according to the tested versions), which I proposed in https://github.com/celery/py-amqp/pull/450 .

auvipy (changes requested):
we will do this after celery 5.6 release cycle is done

PR #457, opened 2026-03-06
Add null check for connection to prevent AttributeError on drain_events

Cyanty:
While null checks exist elsewhere, this specific call was overlooked. This aligns with existing error handling patterns in the codebase.

- reslove #456

### DIFF
--- a/amqp/abstract_channel.py
+++ b/amqp/abstract_channel.py
@@ -96,7 +96,10 @@ class AbstractChannel:
 
         try:
             while not p.ready:
-                self.connection.drain_events(timeout=timeout)
+                conn = self.connection
+                if conn is None:
+                    raise RecoverableConnectionError('Connection already closed')
+                conn.drain_events(timeout=timeout)
 
             if p.value:
                 args, kwargs = p.value
--- a/amqp/abstract_channel.py
+++ b/amqp/abstract_channel.py
@@ -98,7 +98,7 @@ class AbstractChannel:
             while not p.ready:
                 conn = self.connection
                 if conn is None:
-                    raise RecoverableConnectionError('Connection already closed')
+                    raise RecoverableConnectionError('connection already closed')
                 conn.drain_events(timeout=timeout)
 
             if p.value:
--- a/t/unit/test_abstract_channel.py
+++ b/t/unit/test_abstract_channel.py
@@ -83,6 +83,13 @@ class test_AbstractChannel:
             self.c.wait([(50, 61)])
             assert self.c._pending[(50, 61)] is prev
 
+    def test_wait__no_connection(self):
+        self.c.connection = None
+        with pytest.raises(RecoverableConnectionError,
+                           match='Connection already closed'):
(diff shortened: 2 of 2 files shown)


auvipy (changes requested):
can you please add proper unit tests and possibly Integration tests for this proposed change?

ChickenBenny:
Hi @Cyanty, thanks for looking into this and providing a quick fix! 

The local variable assignment `conn = self.connection` before the null check is a solid defensive practice here, and throwing a `RecoverableConnectionError` will definitely stop the ungraceful crash during shutdown.

However, I have a couple of thoughts on this approach:

**1. Is this treating the symptom rather than the root cause?**
Looking at the traceback in #456, the paradox is that a channel actively being closed (`channel.close()`) triggers `_on_close()`, which in turn fires `_do_revive()`. 
If a channel is being intentionally shut down by the worker, should it even be attempting to revive itself? While this `None` check protects the `wait()` loop, it feels like the state logic in `channel.py` (knowing whether the close is intentional vs. an unexpected disconnect) might be the underlying bug. 

**2. Missing test coverage (Agreed with Copilot)**
We absolutely need a unit test for this new exception path. Please add a test case in `t/unit/test_abstract_channel.py` simulating `wait()` when `self.connection` is `None` to ensure it raises `RecoverableConnectionError`.

I'm fine with merging this as a quick-fix / safeguard, but I'd love to hear your thoughts on whether we should also address why `_do_revive()` is being called during a graceful shutdown.

ChickenBenny:
Here is a step-by-step breakdown of my investigation into this issue:

1. **Initiate Shutdown (Celery)**: The Celery worker triggers `shutdown(warm=True)` to exit gracefully.
2. **Close Connection (Kombu)**: Kombu executes `connection.close()` to terminate the connection to RabbitMQ.
3. **Close Channel (py-amqp)**: The underlying `amqp.channel` calls `close()` to send a close frame to the server.
4. **Misinterpretation (py-amqp)**: During this closing phase, the `_on_close` callback is triggered. Because the channel has `auto_recover` enabled, it mistakenly assumes this is an accidental network drop rather than an intentional shutdown.
5. **Attempted Revival (py-amqp)**: The channel tries to rescue itself by calling `_do_revive()`, which attempts to `open()` the channel again and enters a `wait()` loop for the server's response.
6. **Missing Connection & Crash**: Inside the `wait()` loop, it calls `self.connection.drain_events()`. However, because the system is in the middle of shutting down, the upper layers have already cleared the `connection` object and set it to `None`.
7. **Exception Thrown**: Executing `None.drain_events()` results in the fatal `AttributeError: 'NoneType' object has no attribute 'drain_events'`.

I'm not entirely sure whether it is better to fix the root cause at **Step 4** (preventing `_do_revive` during an intentional shutdown) or to keep the defensive safeguard at **Step 6** (the current PR's approach).

auvipy:
I think we can accept both?

auvipy (approved):

PR #458, opened 2026-03-08
Fix prevent channel revival during intentional close to fix shutdown crash

ChickenBenny:
## Description
This PR addresses the root cause of the `AttributeError: 'NoneType' object has no attribute 'drain_events'` during worker shutdown, as reported in #456.

While #457 adds a defensive null check in the `wait()` loop, this PR fixes the underlying state machine issue where an actively closing channel erroneously attempts to auto-recover itself.

## Root Cause
During a graceful shutdown:

channel.close() is called, which explicitly sets `self.is_closing = True`.

While waiting for the `Channel.CloseOk` frame, the `_on_close` callback might be triggered.

Inside `_on_close`, the guard condition only checked if the connection was closing (`self.connection.is_closing`), but completely ignored the channel's own closing state (`self.is_closing`).

As a result, the channel misinterpreted the intentional shutdown as an unexpected drop and called `_do_revive()`, which eventually crashed when it tried to call `drain_events()` on a destroyed connection.

Proposed Changes
Updated the guard in `amqp/channel.py` -> `_on_close()` to check the channel's local state:

```python
# amqp/channel.py
- if not self.connection.is_closing:
+ if not self.connection.is_closing and not self.is_closing:
      self._do_revive()
By adding not self.is_closing, we ensure that if the channel is already in the process of an intentional shutdown, it will gracefully skip the _do_revive() auto-recovery sequence.
```

Related Issues
- Fixes #456
- Alternative/Complementary to #457 (This fixes the root cause, while #457 acts as a defensive safeguard).

### DIFF
--- a/amqp/channel.py
+++ b/amqp/channel.py
@@ -288,7 +288,7 @@ class Channel(AbstractChannel):
                 is the ID of the method.
         """
         self.send_method(spec.Channel.CloseOk)
-        if not self.connection.is_closing:
+        if not self.connection.is_closing and not self.is_closing:
             self._do_revive()
             raise error_for_code(
                 reply_code, reply_text, (class_id, method_id), ChannelError,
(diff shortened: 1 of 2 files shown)


auvipy (changes requested):
can you please check the failing test? also are the tests duplicate?

auvipy (approved):

PR #468, opened 2026-09-07
Drop unused python version check

kajinamit:
pypy-3.9 was removed by f778218d788477539cad672ea678081cd7b9ea0a .


.github/workflows/ci.yaml
```diff
@@ -45,7 +45,6 @@ jobs:
         run: |
           CELERY_ENABLE_SPEEDUPS=1 python setup.py develop
           tox -v -e ${{ matrix.python-version }}-integration-rabbitmq -- -v
-        if: ${{ matrix.python-version != 'pypy-3.9'}}
```
auvipy, line 48: thanks for the patch, what about running pypy 3.11? or it is already running? may be we can explore that in another PR

auvipy (approved):

PR #469, opened 2026-09-15
Decode AMQP field-table type 's' as signed 16-bit

cocofabio:
RabbitMQ and its clients (pika, pamqp/aio-pika, the Java and .NET clients) use field-table type `s` for a signed 16-bit int, per the [AMQP 0-9-1 errata](https://www.rabbitmq.com/amqp-0-9-1-errata#section_3). py-amqp reads it as a length-prefixed short string, so a header like `{'key': 128}` published from aio-pika arrives as `b's\x00\x80'` and decodes to an empty string with a desynced offset — the `IndexError` / `UnicodeDecodeError` / `UnexpectedFrame` reported in celery/kombu#2354.

This decodes `s` as `>h`, same as #465 did for `l`. py-amqp's own writer never emits `s` (ints go out as `I`/`L`), so roundtrips are unaffected.

Repro without a broker:

```python
>>> from pamqp import encode
>>> from amqp.serialization import _read_item
>>> _read_item(b'F' + encode.field_table({'key': 128}), 0)[0]
IndexError: index out of range   # before
{'key': 128}                     # after
```

Tests:
`python -m pytest t/unit` (326 passed on 3.12 and 3.14); `flake8` only reports the two pre-existing warnings on untouched lines.

Root cause of celery/kombu#2354.


auvipy (changes requested):
should we also consider adding integration test for this?

cocofabio:
Added one: `test_publish_get__short_int_header` publishes a header table with the value typed `s` (py-amqp's writer never emits it, so the test hand-crafts the properties the way pika/pamqp/Java send them) and reads it back with `basic_get`. On main it fails with `IndexError` followed by `UnexpectedFrame: Received frame 3 while expecting type: 2`; passes with the fix. Ran it locally against RabbitMQ 4.3.6 — the CI run needs approval.

auvipy (approved):

PR #472, opened 2026-09-20
Fix signedness of byte fields in AMQP tables

sueun-dev:
AMQP field-table types `b` and `B` use the opposite `struct` formats in `_read_item`. A signed `b` value of -1 decodes as 255, while an unsigned `B` value of 255 decodes as -1. [RabbitMQ's encoder](https://github.com/rabbitmq/rabbitmq-server/blob/169a13dc8a86383ca781a3d0be80acef461d0692/deps/rabbit_common/src/rabbit_binary_generator.erl#L119-L128) uses `b` for signed bytes and `B` for unsigned bytes. pamqp also emits `b` for small negative integers, so this changes received header values.

Use `>b` for `b` and `>B` for `B`. The regressions cover both byte boundaries, nested tables, arrays, and preservation of the following field and final offset.

Validated on Python 3.12.14 / Linux arm64:

- Final serialization tests on the original code: 8 failed, 37 passed. Patched: 45 passed.
- `python -m pytest -q`: 372 passed, 25 broker-gated tests skipped.
- Plain-AMQP RabbitMQ tests: 11 passed, 14 TLS cases deselected.
- A separate RabbitMQ 4.3.6 probe published and retrieved all 512 signed/unsigned byte values: 256 mismatches before, zero after. Signed properties came from pamqp 4.0.1; unsigned properties used an explicit `B` wire fixture.
- With Cython speedups: 37 codec tests and the same 11 broker tests passed; the 512-value broker probe also passed. Source distribution and wheel builds passed.

Limitations: TLS, other Python/OS combinations and documentation checks were not run. Flake8 reports the same 39 diagnostics as the base after accounting for shifted line numbers. The pinned pydocstyle 1.1.1 cannot import `collections.Set` on this interpreter. Full compiled unit-test collection fails on an unexported `IGNORED_METHOD_DURING_CHANNEL_CLOSE`; three GenericContent tests also access attributes hidden by Cython. Those compiled-test failures reproduce on the base. This does not claim a passing full compiled unit suite.

Prepared with AI assistance; the checks above were executed locally.


auvipy (commented):
can you also add integration test for this change, please?

auvipy (approved):

PR #473, opened 2026-09-22
Decode AMQP field-table types 'b' and 'B' with the right signedness

cocofabio:
The two branches have their struct formats swapped relative to their own comments: `'b'`, the signed octet, is read with `'>B'`, and `'B'`, the unsigned one, with `'>b'`. The [0-9-1 errata](https://www.rabbitmq.com/amqp-0-9-1-errata#section_3) lists them as signed and unsigned 8-bit respectively.

So a negative integer header from a pika/pamqp/Java publisher arrives as 128..255, and an unsigned `'B'` above 127 arrives negative:

```python
>>> from pamqp import encode
>>> from amqp.serialization import _read_item
>>> _read_item(b'F' + encode.field_table({'k': -1}), 0)[0]
{'k': 255}   # before
{'k': -1}    # after
```

Same shape as #469, and as there py-amqp's own writer emits neither type, so roundtrips are unaffected. The existing `'b'`/`'B'` test params only covered 0..127, where the two formats agree.

Tests:
`python -m pytest t/unit` (327 passed on 3.12 and 3.14) and `t/integration/test_rmq.py -k plain` against RabbitMQ 4.3.6 (12 passed); both new cases fail on main. `flake8 amqp t` shows no warnings beyond the ones already on main.


auvipy (commented):
just accepting the integration test from this pr as the initial pr for fixing this should be merged

auvipy (approved):

